PR #38, opened 2021-05-06
Deposit collateralization threshold level with 2 steps update approach

dimpar:
Risk Manager should act as a last resort for opening an auction to buy a deposit in liquidation. Governance will be able to update the threshold level in 2 steps. First step is to announce that the update is on the way with a new value, and the second step is to finalize a new value after delay time has passed. If the deposit collateralization is higher than the threshold, then an auction will not be opened.

depends on https://github.com/keep-network/coverage-pools/pull/26


contracts/RiskManagerV1.sol
```diff
@@ -126,6 +135,15 @@ contract RiskManagerV1 is Auctioneer, Ownable {
...
     }
 
+    /// @notice Updates collateralization threshold level.
+    /// @dev Can be called by the owner only.
+    /// @param _collateralizationThreshold New collateralization threshold level in percent.
+    function updateCollateralizationThreshold(
+        uint256 _collateralizationThreshold
+    ) external onlyOwner {
+        collateralizationThreshold = _collateralizationThreshold;
+    }
```
pdyraga, line 145: This value needs to be updatable in two steps with a governance delay. Is this something we want to add in this PR or open another one?

contracts/RiskManagerV1.sol
```diff
@@ -18,15 +19,23 @@ interface IDeposit {
...
     using SafeMath for uint256;
 
     uint256 public constant DEPOSIT_LIQUIDATION_IN_PROGRESS_STATE = 10;
     uint256 public constant DEPOSIT_LIQUIDATED_STATE = 11;
+    // Auction will not be opened if the deposit collateralization level does not
+    // fall below this collateralization threshold.
+    // Risk manager should open a coverage pool auction for only those deposits
+    // that nobody else is willing to purchase. The default value can be updated
+    // by the governance at any moment.
+    uint256 public collateralizationThreshold = 101; // percent
```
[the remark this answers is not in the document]
pdyraga: That's the crux. Let's create a sheet and try to model different options. This value can be updated by the governance by we need to help them analyze and understand how does it work and assign good default value. Two cases are especially interesting:
- what if the gas price is so high that although taking deposit is theoretically profitable, no one wants to do that?
- what if the collateralization level goes up quickly after the risk manager decides to open an auction?

https://docs.keep.network/tbtc/#_undercollateralization

test/RiskManagerV1.test.js
```diff
@@ -71,28 +72,47 @@ describe("RiskManagerV1", () => {
...
       })
+    })
 
-      it("should emit NotifiedLiquidation event", async () => {
-        await expect(notifyLiquidationTx)
-          .to.emit(riskManagerV1, "NotifiedLiquidation")
-          .withArgs(mockIDeposit.address, notifier.address)
-      })
+    context(
+      "when deposit is in liquidation state and not above collateralization threshold",
```
pdyraga, line 92: `s/and not above collateralization threshold/and at the collateralization threshold` maybe?

contracts/RiskManagerV1.sol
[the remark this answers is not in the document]
dimpar: > what if the gas price is so high that although taking deposit is theoretically profitable, no one wants to do that?

Here's a [sheet](https://docs.google.com/spreadsheets/d/18tDejlULfTN4RG_xgfpcG-K5G8aOwhM6kjlQnPiC0Z4/edit?usp=sharing) based on today's BTC/ETH ratio `0.078`. If my understanding is right, then threshold `101%` is not profitable when the gas price is between 875 - 900. Everything below this gas price should be profitable to purchase a deposit outside the coverage pool.

> what if the collateralization level goes up quickly after the risk manager decides to open an auction?

I'd need some help on that one. If collateralization level was low enough to open an auction in coverage pool and then it went up above the threshold, I'm not sure what should happen in this case.  We have a function to early close an auction, but not sure if this is what we want to trigger?

contracts/RiskManagerV1.sol
dimpar: I started adding it in this PR. What should be the governance delay in this case, is it the same as for auction length? There might be a situation when BTC/ETH ratio changes very quickly and not sure if 12 hours is okay to change collateralization threshold. Maybe we should allow the governance to react faster than 12 hours? BTC/ETH ratio is one of the factors that tells us when it is still profitable to buy a deposit outside this coverage pool.

contracts/RiskManagerV1.sol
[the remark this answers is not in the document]
dimpar: In the sheet above, I've considered two scenarios based on 1 BTC as a deposit and the historic ETH/BTC ratios from May '20 - May '21. In May '20 it was `~0.03`, now it is `~0.075`. If we consider fluctuations in % for the average gas price of `125 gwei` then it varies between `100-101`.  At `100%` collateralization is already not profitable to buy, but at `101%` still makes sense to buy outside the coverage pool. So maybe we should change the default value to `100`, wdyt @pdyraga ?

contracts/RiskManagerV1.sol
dimpar: 404eebe

test/RiskManagerV1.test.js
dimpar: 404eebe

contracts/RiskManagerV1.sol
```diff
@@ -172,6 +223,21 @@ contract RiskManagerV1 is Auctioneer, Ownable {
...
+    /// @return Remaining time in seconds.
+    function getRemainingCollateralizationThresholdUpdateTime()
+        external
+        view
+        returns (uint256)
+    {
+        return
+            getRemainingChangeTime(
+                collateralizationThresholdInitiated,
+                GOVERNANCE_TIME_DELAY
```
tomaszslabon, line 237: We have multiple governance delay constants. Perhaps we could name it:
`COLLATERALIZATION_THRESHOLD_GOVERNANCE_DELAY` ?

Also, if I remember correctly, it was suggested to keep governance-related consts in `CoveragePoolConstants`.
Am I right, @pdyraga?

contracts/RiskManagerV1.sol
```diff
@@ -45,6 +47,16 @@ contract RiskManagerV1 is Auctioneer, Ownable {
...
+    // Auction will not be opened if the deposit collateralization level does not
+    // fall below this collateralization threshold.
+    // Risk manager should open a coverage pool auction for only those deposits
+    // that nobody else is willing to purchase. The default value can be updated
+    // by the governance in two steps. First step is to begin the update process
+    // with the new value and the second step is to finalize it after
+    // GOVERNANCE_TIME_DELAY has passed.
+    uint256 public collateralizationThreshold = 101; // percent
+    uint256 public newCollateralizationThreshold;
+    uint256 public collateralizationThresholdInitiated;
```
pdyraga, line 59: We have `auctionLengthChangeInitiated` so this one should be `collateralizationThresholdChangeInitiated`.

contracts/RiskManagerV1.sol
```diff
@@ -142,6 +161,38 @@ contract RiskManagerV1 is Auctioneer, Ownable {
         earlyCloseAuction(auction);
     }
 
+    /// @notice Begins the collateralization threshold update process.
+    /// @dev Can be called only by the contract owner. The collateralization
+    ///      threshold should be adjusted by taking into account factors like
+    ///      collateralization ratio and gas price.
```
pdyraga, line 167: Instead of saying 
> ... factors like collateralization ratio and gas price.

I would go a little more specific and say:
> ... factors like, for example, tBTC deposit undercollateralized threshold percent, tBTC deposit severely undercollateralized threshold percent, and the gas price.

contracts/RiskManagerV1.sol
```diff
@@ -142,6 +161,38 @@ contract RiskManagerV1 is Auctioneer, Ownable {
...
+    /// @param _collateralizationThreshold New collateralization threshold in percent.
+    function beginCollateralizationThresholdUpdate(
+        uint256 _collateralizationThreshold
```
pdyraga, line 170: Maybe `_newCollateralizationThreshold` to stay consistent with `beginAuctionLengthUpdate`?

contracts/RiskManagerV1.sol
pdyraga: I think `GOVERNANCE_TIME_DELAY` is fine in this case because all risk manager governable parameters will have [the same delay of 12 hours](https://github.com/keep-network/coverage-pools/blob/main/docs/design.adoc#governance). If they would have different values, then sure, using different names would make sense.

> Also, if I remember correctly, it was suggested to keep governance-related consts in `CoveragePoolConstants`.

Only if they belong to coverage pools world, Risk Manager is a kind of a plug-in in this system, so I think what's been proposed here and earlier in the auction length update PR makes sense.

contracts/RiskManagerV1.sol
```diff
@@ -45,6 +47,16 @@ contract RiskManagerV1 is Auctioneer, Ownable {
...
     uint256 public constant DEPOSIT_LIQUIDATION_IN_PROGRESS_STATE = 10;
     uint256 public constant DEPOSIT_LIQUIDATED_STATE = 11;
+    // Auction will not be opened if the deposit collateralization level does not
+    // fall below this collateralization threshold.
+    // Risk manager should open a coverage pool auction for only those deposits
+    // that nobody else is willing to purchase. The default value can be updated
+    // by the governance in two steps. First step is to begin the update process
+    // with the new value and the second step is to finalize it after
+    // GOVERNANCE_TIME_DELAY has passed.
+    uint256 public collateralizationThreshold = 101; // percent
```
pdyraga, line 57: I'd consider making it a constructor parameter, just like `_auctionLength` for consistency. It should also give us a chance to adjust this value last minute before the deployment without having to change the code.

test/RiskManagerV1.test.js
```diff
@@ -248,11 +268,138 @@ describe("RiskManagerV1", () => {
...
+        await expect(
+          riskManagerV1
+            .connect(notifier)
+            .beginCollateralizationThresholdUpdate(102)
+        ).to.be.revertedWith("Ownable: caller is not the owner")
+      })
+    })
+  })
+
+  describe("finalizeAuctionLengthUpdate", () => {
```
pdyraga, line 315: `finalizeCollateralizationThresholdUpdate`

contracts/RiskManagerV1.sol
pdyraga: I do not think the current ETH:BTC price ratio affects this threshold that much. 🤔 But let's discuss it separately in the thread about the correct initial value.

contracts/RiskManagerV1.sol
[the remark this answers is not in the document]
pdyraga: I'd recommend opening an issue with those findings plus adding a backlog item to find the correct initial value so that we do not have to block this PR and having this discussion there. WDYT?

test/RiskManagerV1.test.js
dimpar: 81c3fe3

contracts/RiskManagerV1.sol
dimpar: 81c3fe3

contracts/RiskManagerV1.sol
dimpar: 81c3fe3

contracts/RiskManagerV1.sol
dimpar: 81c3fe3

contracts/RiskManagerV1.sol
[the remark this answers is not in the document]
dimpar: Sure. [issue](https://github.com/keep-network/coverage-pools/issues/64) & backlog item was also added.

pdyraga (approved):
Looks good to me, leaving it in @tomaszslabon's hands.

tomaszslabon (approved):

PR #39, opened 2021-05-06
Signer bonds escrow

lukasz-zimnoch:
Depends on: #26 

ETH purchased by the risk manager from tBTC signer bonds needs to be
swapped and deposited back to the coverage pool as collateral.
In the case it can not be done automatically, the governance has
the power to ask the risk manager to deposit ETH from purchased
signer bonds into an escrow the governance can later withdraw from
and do the swap manually. `SignerBondsEscrow` is a simple escrow
implementation allowing the risk manager to store purchased ETH
signer bonds so that governance can later swap them manually and
deposit as coverage pool collateral.

PR #43, opened 2021-05-07
Reuse TBTC from early closed auctions

lukasz-zimnoch:
Depends on: #39 

tBTC from early closed auctions should be reused if possible. This pull request provides such functionality by introducing a tBTC surplus mechanism in the `RiskManagerV1` contract. Once an auction is early closed, its transferred amount is added to the surplus pool. When a new auction is created, the underlying deposit lot size is checked against the current value of the tBTC surplus pool. In case the surplus is big enough to cover the entire deposit lot size, the deposit is liquidated directly, without performing an auction.


pdyraga:
After giving it more thought, I think this mechanism - although logically sound - is risky, a little too complex (read: expensive), and we could simplify it. My two concerns are:
- We need another mapping, we need to update and calculate reservations. It is relatively complex and it is harder to grasp what is happening, plus, the execution costs more gas. 
- There are corner cases when, for example, the surplus is 0.9999 TBTC and we would open an auction for 0.0001 TBTC. This is less than the minimum tBTC lot size, and the governance can not predict all those cases, thus, the auction length would have to be somehow adjusted which makes the mechanism even harder.

I would consider changing this mechanism so that it keeps bookkeeping the surplus but instead of reducing the auction value, it would go straight to purchasing signer bonds if the surplus value allows for it. For example:
- liquidation of 1 TBTC deposit, auction opened for 1 TBTC and early closed, 0.3 TBTC goes to the auctioneer.
- liquidation of 1 TBTC deposit, auction opened for 1 TBTC and early closed, 0.8 TBTC goes to the auctioneer.
- liquidation of 1 TBTC deposit, there is 1.1 TBTC in the surplus, instead of opening an auction, we purchase signer bonds and reduce the surplus to 0.1 TBTC.

Please give it some thought, there is a chance some gremlins are hidden in this approach.

lukasz-zimnoch:
> After giving it more thought, I think this mechanism - although logically sound - is risky, a little too complex (read: expensive), and we could simplify it. My two concerns are:
> 
>     * We need another mapping, we need to update and calculate reservations. It is relatively complex and it is harder to grasp what is happening, plus, the execution costs more gas.
> 
>     * There are corner cases when, for example, the surplus is 0.9999 TBTC and we would open an auction for 0.0001 TBTC. This is less than the minimum tBTC lot size, and the governance can not predict all those cases, thus, the auction length would have to be somehow adjusted which makes the mechanism even harder.
> 
> 
> I would consider changing this mechanism so that it keeps bookkeeping the surplus but instead of reducing the auction value, it would go straight to purchasing signer bonds if the surplus value allows for it. For example:
> 
>     * liquidation of 1 TBTC deposit, auction opened for 1 TBTC and early closed, 0.3 TBTC goes to the auctioneer.
> 
>     * liquidation of 1 TBTC deposit, auction opened for 1 TBTC and early closed, 0.8 TBTC goes to the auctioneer.
> 
>     * liquidation of 1 TBTC deposit, there is 1.1 TBTC in the surplus, instead of opening an auction, we purchase signer bonds and reduce the surplus to 0.1 TBTC.
> 
> 
> Please give it some thought, there is a chance some gremlins are hidden in this approach.

You're right. I had the same feelings while implementing that approach. Please see https://github.com/keep-network/coverage-pools/pull/43/commits/259f7e82bc03c1393c1454c1ac12fb46b0831b39. I made some change towards a simpler approach.

contracts/Auction.sol
```diff
@@ -139,6 +141,14 @@ contract Auction is IAuction {
         return self.amountOutstanding;
     }
 
+    function amountTransferred() external view returns (uint256) {
+        // Need to check whether the auction is open as harikari is done
+        // upon auction close so there is no possibility to determine
+        // the transferred amount on a closed auction.
+        require(self.amountOutstanding > 0, "Auction must be open");
```
pdyraga, line 148: I am not sure if I follow. `selfdestruct` should destroy the contract so that it cannot be called in the future. This is only a problem in the *current* transaction and if I understand correctly, we want to make sure functions are called in the right order in `Auctioneer.earlyClose` or in the code calling `Auctioneer.earlyClose`, so that the amount transferred is evaluated before `Auction.earlyClose`.

If so, why don't we return `surplus = amountInitial - amountOutstanding` from `earlyClose`? We do not need this function then, the risk is gone and we consume less gas. This would also play well with another comment I left in `RiskMangerV1` about having the surplus returned from `Auctioneer.earlyClose`. WDYT?

contracts/Auction.sol
```diff
@@ -32,6 +32,7 @@ contract Auction is IAuction {
         Auctioneer auctioneer;
         // the auction price, denominated in tokenAccepted
         uint256 amountOutstanding;
+        uint256 amountInitial;
```
pdyraga, line 35: Maybe `amountDesired`? The desired amount does not change, only the outstanding amount does.

test/RiskManagerV1.test.js
```diff
@@ -69,28 +73,181 @@ describe("RiskManagerV1", () => {
...
+    context(
+      "when deposit is in liquidation state and the surplus pool is " +
+        "smaller than the deposit lot size",
+      () => {
+        let notifyLiquidationTx
+        let auctionAddress
+
+        beforeEach(async () => {
+          const surplus = to1ePrecision(30, 16)
+          await testToken.mint(owner.address, surplus)
```
pdyraga, line 116: We should rename `testToken` into `tbtcToken` at some point. It is confusing now.

test/RiskManagerV1.test.js
```diff
@@ -69,28 +73,181 @@ describe("RiskManagerV1", () => {
       })
     })
 
-    context("when deposit is in liquidation state", () => {
-      let notifyLiquidationTx
+    context(
+      "when deposit is in liquidation state and the surplus pool is empty",
```
pdyraga, line 77: We could also have:
```
context("when deposit is in liquidation state", () => {
  it("should emit NotifiedLiquidation event", async () => {...}))
  context("when the surplus pool is empty, () => {...})
  context("when the surplus pool is smaller then the deposit lot size, () => {...})
  // (...)
})
```

contracts/RiskManagerV1.sol
```diff
@@ -131,12 +141,11 @@ contract RiskManagerV1 is Auctioneer, Ownable {
...
         emit NotifiedLiquidated(depositAddress, msg.sender);
 
-        // TODO: In case of an auction early close, we might end up having
-        //       TBTC hanging in this contract. Need to decide what to do with
-        //       these tokens.
-
         Auction auction = Auction(depositToAuction[depositAddress]);
 
+        // Add auction's transferred amount to the surplus pool.
+        tbtcSurplus = tbtcSurplus.add(auction.amountTransferred());
```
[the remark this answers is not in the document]
lukasz-zimnoch: Good point, see https://github.com/keep-network/coverage-pools/pull/43/commits/109989daf5b6ee78f89ee8803f381efb4c32035c

contracts/Auction.sol
lukasz-zimnoch: Done in https://github.com/keep-network/coverage-pools/pull/43/commits/ba3de981619c6aea018188c28b64fc2e99f4ce49

contracts/Auction.sol
lukasz-zimnoch: Also done in https://github.com/keep-network/coverage-pools/pull/43/commits/ba3de981619c6aea018188c28b64fc2e99f4ce49

test/RiskManagerV1.test.js
lukasz-zimnoch: See https://github.com/keep-network/coverage-pools/pull/43/commits/0c6a7abca95a97ff5edd1f9fbab7302481698e8a

test/RiskManagerV1.test.js
lukasz-zimnoch: Improved the structure of the context tree in https://github.com/keep-network/coverage-pools/pull/43/commits/c151f7be59a000a30fc5eadee7ae89944d76fc5a. However, there is no possibility to check `NotifiedLiquidation` in one place as `beforeEach` hooks look slightly different in each case.

lukasz-zimnoch:
@pdyraga, all comments addressed!

contracts/Auction.sol
pdyraga: Right, I missed this fact. Anyway, I'd remove the `require` check from here. The `Auctioneer` needs to fetch all the information it needs before calling `earlyClose()`. This is true for `amountTransferred` but also for any other field or function in `Auction` contract: `amountInitial`, `amountOutstanding`, `startTime`... I do not think we should treat `amountTransferred()` in any special way.

contracts/Auctioneer.sol
```diff
@@ -125,19 +125,24 @@ contract Auctioneer is CloneFactory {
 
     /// @notice Tears down an open auction with given address immediately.
     /// @dev Can be called by contract owner to early close an auction if it
-    ///      is no longer needed.
-    function earlyCloseAuction(Auction auction) internal {
+    ///      is no longer needed. Bear in mind that funds from the early closed
+    ///      auction last on the auctioneer contract. Calling code should take
+    ///      care of them.
+    /// @return Amount of funds transferred to the auction at the time of closing.
```
pdyraga, line 131: More precisely, isn't it "the amount transferred to this contract by the Auction being early closed"?

test/system/liquidation.test.js
```diff
@@ -166,7 +166,7 @@ describeFn("System -- liquidation happy path", () => {
     })
 
     it("should consume a reasonable amount of gas", async () => {
-      await expect(parseInt(tx.gasLimit)).to.be.lessThan(480000)
+      await expect(parseInt(tx.gasLimit)).to.be.lessThan(485000)
```
pdyraga, line 169: Not in this PR but we quickly need to figure out why such high gas limit is needed. Estimations in earlier PRs about purchasing signer bonds were saying about ~250k.

test/RiskManagerV1.test.js
```diff
@@ -43,9 +46,9 @@ describe("RiskManagerV1", () => {
     masterAuction = await Auction.deploy()
     await masterAuction.deployed()
 
-    const RiskManagerV1 = await ethers.getContractFactory("RiskManagerV1")
+    const RiskManagerV1 = await ethers.getContractFactory("RiskManagerV1Stub")
```
pdyraga, line 49: Nitpick: can we `s/RiskMangerV1/RiskManagerV1Stub`? We use this pattern in other test files where we have to work with stubs.

test/system/liquidation.test.js
lukasz-zimnoch: Yep, I will try to figure out this separately.

contracts/Auction.sol
lukasz-zimnoch: https://github.com/keep-network/coverage-pools/pull/43/commits/495e090974a40c594c8304435a477d3c5eab2fe6

contracts/Auctioneer.sol
lukasz-zimnoch: https://github.com/keep-network/coverage-pools/pull/43/commits/495e090974a40c594c8304435a477d3c5eab2fe6

test/RiskManagerV1.test.js
lukasz-zimnoch: https://github.com/keep-network/coverage-pools/pull/43/commits/0fd0f595c291327882884f51b896154a28b23fac

pdyraga (approved):

PR #44, opened 2021-05-07
Governance should be able to approve risk managers

tomaszslabon:
This pull requests:
- adds functions for two-stage approving/unapproving of risk managers:
   - `beginRiskManagerApproval`
   - `finalizeRiskManagerApproval`
   - `beginRiskManagerUnapproval`
   - `finalizeRiskManagerUnapproval`
   
 Before the `begin...` and `finalize..` functions a delay of 30 days must elapse.
 Approved risk managers are stored in a mapping.
 Timestamps of initiation of approval/unapproval of a risk manager are also stored in mappings.
 
 Only approved risk managers can seize funds.

### DIFF
--- a/contracts/CoveragePool.sol
+++ b/contracts/CoveragePool.sol
@@ -12,14 +12,27 @@ import "@openzeppelin/contracts/math/SafeMath.sol";
 contract CoveragePool is Ownable {
     using SafeMath for uint256;
 
+    uint256 public constant RISK_MANAGER_APPROVAL_TIME_DELAY = 2 weeks;
+
     AssetPool public assetPool;
     IERC20 public collateralToken;
 
-    mapping(address => bool) private approvedManagers;
+    // Maps risk managers to the timestamp they are considered approved
+    mapping(address => uint256) public riskManagerApprovalTimestamps;
 
-    /// @notice Throws if called by a Risk Manager that has not been approved
+    /// @notice Throws if called by a risk manager that has not been approved
+    /// or approval time delay has not elapsed
+    /// @dev Risk manager approval delay is already added to approval timestamps
     modifier onlyApprovedRiskManager() {
-        require(approvedManagers[msg.sender], "Risk manager not approved");
+        require(
+            riskManagerApprovalTimestamps[msg.sender] > 0,
+            "Risk manager not approved"
+        );
+        require(
+            /* solhint-disable-next-line not-rely-on-time */
+            block.timestamp >= riskManagerApprovalTimestamps[msg.sender],
+            "Approval delay has not elapsed"
+        );
         _;
     }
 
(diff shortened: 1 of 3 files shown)


contracts/CoveragePool.sol
```diff
@@ -26,8 +41,14 @@ contract CoveragePool is Ownable {
...
-    function approveRiskManager() external onlyOwner {
-        //TODO: implement
+    /// @notice Approves the given risk manager so that it can withdraw funds
+    /// from the asset pool (after approval delay has elapsed)
+    /// @param riskManager Address of a risk manager to be approved
+    function approveRiskManager(address riskManager) external onlyOwner {
+        riskManagerApprovalTimestamps[riskManager] =
+            /* solhint-disable-next-line not-rely-on-time */
+            block.timestamp +
+            RISK_MANAGER_APPROVAL_TIME_DELAY;
```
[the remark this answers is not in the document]
tomaszslabon: I've changed the approval to a two-step process and added approval (also a two-step process).

contracts/CoveragePool.sol
```diff
@@ -26,11 +39,76 @@ contract CoveragePool is Ownable {
         collateralToken = _assetPool.collateralToken();
     }
 
-    function approveRiskManager() external onlyOwner {
-        //TODO: implement
+    /// @notice Begins risk manager approval process
```
pdyraga, line 42: `.` at the end, please. Same for the other `@notice`s below.

contracts/CoveragePool.sol
```diff
@@ -42,14 +120,11 @@ contract CoveragePool is Ownable {
...
     {
-        uint256 FLOATING_POINT_DIVISOR =
-            CoveragePoolConstants.getFloatingPointDivisor();
-
         uint256 amountToSeize =
             collateralToken
                 .balanceOf(address(assetPool))
                 .mul(portionToSeize)
-                .div(FLOATING_POINT_DIVISOR);
+                .div(CoveragePoolConstants.FLOATING_POINT_DIVISOR);
```
pdyraga, line 127: Can we do the same - in a separate PR - in all other places where `getFloatingPointDivisor` is used?

test/CoveragePool.test.js
```diff
@@ -61,20 +61,275 @@ describe("CoveragePool", () => {
     await assetPool.connect(underwriter).deposit(to1e18(400))
   })
 
+  describe("beginRiskManagerApproval", () => {
+    context("when caller is not the governance", () => {
+      it("should revert", async () => {
+        const notGovernance = await ethers.getSigner(5)
```
pdyraga, line 67: Non-blocking, but we could abstract it as e.g. `thirdParty` and keep it next to `riskManager` and `underwriter`, just to make sure signer with index 5 is not going to be accidentally used in the future for some other role.

test/CoveragePool.test.js
```diff
@@ -61,20 +61,275 @@ describe("CoveragePool", () => {
...
+          .connect(governance)
+          .beginRiskManagerApproval(riskManager.address)
+      })
+
+      it("should not approve risk manager", async () => {
+        expect(await coveragePool.approvedRiskManagers(riskManager.address)).to
+          .be.false
+      })
+
+      it("should store risk manager approval timestamp", async () => {
```
pdyraga, line 88: `approval process begin timestamp` perhaps? Same for the other, similar tests.

pdyraga (dismissed):
Left some non-blocking comments, it looks good. 

I would merge it but GitHub Actions are having a partial outage and I can't get a green build here. I am fine with addressing review comments in a separate PR or here, your call @tomaszslabon .

contracts/CoveragePool.sol
tomaszslabon: Sure, I will do that in a separate PR:
https://github.com/keep-network/coverage-pools/pull/54

test/CoveragePool.test.js
tomaszslabon: I changed it to "should store approval process begin timestamp".

pdyraga (approved):
Let's 🚢 it.

PR #52, opened 2021-05-17
Document the v1 coverage pool

pdyraga:
This document describes the v1, single-asset version of a coverage pool:
- components,
- depositing and withdrawal from the pool,
- tBTC v1 risk manager,
- upgradability,
- governance.

v2 of a coverage pool includes multi-asset coverage and rewards and is covered in v2 documentation.
Existing documentation for v2 has been moved to `docs-v2/design.adoc ` and this is also where the work from #2 will eventually land.


dougvk (dismissed):
LGTM

docs/design.adoc
```diff
@@ -0,0 +1,282 @@
...
+
+* ETH is automatically swapped to asset pool underlying token on Uniswap,
+* ETH is deposited in the escrow contract allowing the governance to do the swap 
+  manually and deposit the underlying token to the asset pool.
+
+Which strategy is used is a governable parameter.
+
+In case signer bonds were purchased by a third party before the auction was
+fully filled, TBTC acquired by the risk manager from potential partial auction
+takes will be used in the future, reducing the amount of a next auction.
```
lukasz-zimnoch, line 241: Regarding https://github.com/keep-network/coverage-pools/pull/43#issuecomment-843990253 I think we should change it to: 
```
In case signer bonds were purchased by a third party before the auction was
fully filled, TBTC acquired by the risk manager from potential partial auction
takes will be used in the future, to purchase signer bonds once the gathered surplus value allows for it.
```

docs/design.adoc
```diff
@@ -0,0 +1,282 @@
+= Coverage pool
```
lukasz-zimnoch, line 1: Maybe we can link that doc somewhere in the `README` file?

docs/design.adoc
pdyraga: Please see https://github.com/keep-network/coverage-pools/pull/52/commits/2e778f6cbc55a25b484b825a3cbed90c4b6dda63. There will be some additional edits to this document in the future, especially about the fixed withdrawal period and what happens if someone decides not to withdraw after the cooldown period completes. But all of that in a separate PR.

docs/design.adoc
pdyraga: Done in https://github.com/keep-network/coverage-pools/pull/52/commits/fd0f88ac27910244f04fcf3b5cdad9492e21c554.

tomaszslabon (approved):

PR #53, opened 2021-05-17
System tests of risk manager liquidation flow

lukasz-zimnoch:
Here we introduce a basic system test framework and add a case for deposit liquidation happy path. System tests can be started using the following command:
```
FORKING_URL=<forking-url> yarn test:system
```
where `FORKING_URL` should point to a mainnet node which contains archive data.


pdyraga:
Can we rebase the code so that this feature does not depend on #43? I am afraid it may take us some time to merge #43 and we could have this code earlier in `main`.

lukasz-zimnoch:
> Can we rebase the code so that this feature does not depend on #43? I am afraid it may take us some time to merge #43 and we could have this code earlier in `main`.

Ok, done.

test/system/liquidation.test.js
```diff
@@ -0,0 +1,127 @@
...
+      masterAuction.address,
+      auctionLength
+    )
+    await riskManagerV1.deployed()
+
+    tbtcDeposit = await ethers.getContractAt("IDepositStub", depositAddress)
+
+    bidder = await impersonateAccount(bidderAddress)
+  })
+
```
dimpar, line 77: We can also add validation of the initial params after the setup. It will show that certain params have changed because of ex. calling `takeOffer()`. Something like:

```
describe("initial state", () => {
    it("should assert a deposit is in active state", async () => {
      expect(await tbtcDeposit.currentState()).to.equal(5) // Active
    })

    it("should assert an auction does not exist", async () => {
      const auctionAddress = await riskManagerV1.depositToAuction(
        tbtcDeposit.address
      )
      // auctionAddress == ZERO_ADDRESS
    })
  })
  ```

But this is a nice to have I guess..

**Comment by dimpar** [comment]
This is cool, I didn't realize we can do the mainnet forking and write tests against it :)

File: test/system/liquidation.test.js [hunk 2]
```diff
@@ -0,0 +1,127 @@
...
+// deposit (https://allthekeeps.com/deposit/0x55d8b1dd88e60d12c81b5479186c15d07555db9d)
+// which is ready to be liquidated at the starting block. The bidder which
+// takes the offer is also a real account with actual tBTC balance. At the
+// end of the scenario, the risk manager should liquidate the deposit successfully,
+// and 66% of the deposit's bonded amount should land on the signer processor.
+describeFn("System -- liquidation happy path", () => {
+  const startingBlock = 12368838
+  const tbtcTokenAddress = "0x8daebade922df735c38c80c7ebd708af50815faa"
+  const depositAddress = "0x55d8b1dd88e60d12c81b5479186c15d07555db9d"
+  const bidderAddress = "0xa0216ED2202459068a750bDf74063f677613DA34"
```
pdyraga, line 26: Why this particular address? Because it has some TBTC and ETH at the starting block?

test/system/liquidation.test.js
```diff
@@ -0,0 +1,127 @@
...
+    await signerBondsSwapStrategy.deployed()
+
+    const Auction = await ethers.getContractFactory("Auction")
+
+    const masterAuction = await Auction.deploy()
+    await masterAuction.deployed()
+
+    const CoveragePoolStub = await ethers.getContractFactory("CoveragePoolStub")
+    coveragePool = await CoveragePoolStub.deploy()
+    await coveragePool.deployed()
```
pdyraga, line 61: Why do we deploy a stub? I'd rather deploy real `CoveragePool` for a system test.

test/system/liquidation.test.js
```diff
@@ -0,0 +1,127 @@
...
+
+      // Deposit has been liquidated.
+      expect(await tbtcDeposit.currentState()).to.equal(11) // LIQUIDATED
+    })
+
+    it("should swap signer bonds", async () => {
+      // No funds should last on the risk manager contract.
+      await expect(tx).to.changeEtherBalance(riskManagerV1, 0)
+
+      // All funds should be moved to the signer bonds processor contract.
```
pdyraga, line 113: s/signer bonds processor contract/signer bonds swap strategy

test/system/liquidation.test.js
```diff
@@ -0,0 +1,127 @@
...
+      await expect(tx).to.changeEtherBalance(
+        signerBondsSwapStrategy,
+        bondedAmount.mul(bondedAmountPercentage).div(100)
+      )
+    })
+
+    it("should consume a reasonable amount of gas", async () => {
+      await expect(parseInt(tx.gasLimit)).to.be.lessThan(435000)
```
pdyraga, line 121: That feels... high. 🙀  The estimation from https://github.com/keep-network/coverage-pools/pull/26 was that it should cost about 89301 + 125278. 
 Is this finally reduced to 215k because we call `selfdestruct` on the contract? How would that look like without the `selfdestruct`? Gas refunds for SSTORE and SELFDESTRUCT can be removed at some point: https://eips.ethereum.org/EIPS/eip-3298.

cc @dimpar

package.json
```diff
@@ -13,19 +15,18 @@
...
     "solhint": "^3.3.4",
-    "typescript": "^3.7.0",
-    "solhint-config-keep": "github:keep-network/solhint-config-keep"
-  },
-  "dependencies": {
-    "@openzeppelin/contracts": "^3.3.0"
+    "solhint-config-keep": "github:keep-network/solhint-config-keep",
+    "typescript": "^3.7.0"
   },
+  "name": "@keep-network/coverage-pool",
```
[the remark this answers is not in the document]
lukasz-zimnoch: `pretty-format-json` pre-commit hook is rather against this idea 😞

hardhat.config.js
```diff
@@ -5,4 +5,18 @@ module.exports = {
...
     version: "0.7.6",
   },
+
+  networks: {
+    hardhat: {
+      forking: {
+        // forking is enabled only if FORKING_URL env is provided
+        enabled: !!process.env.FORKING_URL,
+        // URL should point to a node with archival data (Alchemy recommended)
+        url: process.env.FORKING_URL || "",
```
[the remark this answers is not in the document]
lukasz-zimnoch: Hardhat expects either a valid URL or a blank string. Neither `undefined` nor `null` work here.

test/system/liquidation.test.js
lukasz-zimnoch: Yep, `depositAddress` is chosen to be an address of a liquidation-ready deposit. `bidderAddress`, in turn, must have enough TBTCs to buy signer bonds and perform the liquidaton.

test/system/liquidation.test.js
lukasz-zimnoch: I copy-pasted this part from integration tests and forget to change it. I will fix that.

test/system/liquidation.test.js
lukasz-zimnoch: Yep, but estimations from #26 didn't include real-world tBTC overhead. Regarding gas refund and `selfdestruct`, I made an experiment by commenting it out and... the gas used is almost the same. It seems that `delete self;` makes the real difference as without it, gas usage is around ~40k bigger.

dimpar: The latest estimation of `takeOffer()` which also closes an auction and purchases signer bonds is about `209840` https://github.com/keep-network/coverage-pools/pull/38#discussion_r635314978
When running tests in tBTC and trying to estimate `purchaseSignerBondsAtAuction` the gas usage actually varied from around `100k` to `145k` that's why put an avg. of `125k`.
Anyway, yes, `435k` looks like almost doubles the avg. gas usage for `takeOffer()` running a full cycle with purchasing bonds.

pdyraga: What if we decrease the gas limit to e.g. 220k? Will the transaction pass? That 435k gas limit could be a result of not very precise estimation. But it could also be the real amount of gas burned before some data is deleted and gas is refunded. In this case, the bidder needs to have enough ETH to cover 435k gas.

lukasz-zimnoch: Transaction reverts with a gas limit equal to 220k. Seems it burns around 435k and then gets a gas refund due to `delete self;`.

test/system/liquidation.test.js
lukasz-zimnoch: Please see https://github.com/keep-network/coverage-pools/pull/53/commits/2c6ccdba064b98e9a2115957288c9007456156a1

contracts/test/DepositStub.sol
```diff
@@ -6,10 +6,16 @@ import "@openzeppelin/contracts/token/ERC20/SafeERC20.sol";
 
 import "../RiskManagerV1.sol";
 
+/// @dev Stub IDeposit interface which adds method needed by tests but not
+///      necessarily by the production code.
+interface IDepositStub is IDeposit {
+    function notifyRedemptionSignatureTimedOut() external;
+}
+
```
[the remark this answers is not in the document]
lukasz-zimnoch: See https://github.com/keep-network/coverage-pools/pull/53/commits/a9f8df1863eb24053beb142b89c052c21d763f94

test/helpers/contract-test-helpers.js
```diff
@@ -37,30 +37,47 @@ async function increaseTime(time) {
...
+    })
+  }
+
+  // Return the account's signer.
+  return await ethers.getSigner(accountAddress)
+}
 
-  // Return the contract account's signer.
-  return await ethers.getSigner(contractAddress)
+async function resetFork(blockNumber) {
```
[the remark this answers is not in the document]
lukasz-zimnoch: Done in https://github.com/keep-network/coverage-pools/pull/53/commits/a359b060e67aff8d189738b3ab59aa97a8392bdb

test/system/liquidation.test.js
lukasz-zimnoch: Fixed in https://github.com/keep-network/coverage-pools/pull/53/commits/8a07c78529c05ba9e3d4a83afce054a726142d76

test/system/liquidation.test.js
lukasz-zimnoch: Done in https://github.com/keep-network/coverage-pools/pull/53/commits/75376a9a2fb20fbe099174b2bb224fa34c48f7c7

lukasz-zimnoch:
> can we change it to a value closer to the real world scenario? Ex. 100?

Not in this case. This scenario uses https://allthekeeps.com/deposit/0x55d8b1dd88e60d12c81b5479186c15d07555db9d which has collateralization above 400% and was liquidated due to signer timeout.

dimpar:
While changing the collateralization threshold level Łukasz & I discovered an issue how the percentage is being calculated when the deposit is under liquidation state. It seems that the bonds are taken in liquidation state, so the collateralaiztion threshold is going to be 0. This issue is outside the scope of these tests and will be tackled in a separate PR.

dimpar (approved):

PR #55, opened 2021-05-19
Risk Manager should be able to refuse takeOffer calls on the Auction for Deposit liquidated in the meantime

tomaszslabon:
This pull request:
- adds a check (`onAuctionPartiallyFilled`) for the state of the deposit associated with an auction in `Auctioneer`'s `offerTaken` function.

Note: As the mapping between auctions and deposits is stored in `RiskManagerV1`, it cannot be accessed directly from the `Auctioneer` contract. `Auctioneer` calls function `onAuctionPartiallyFilled` that is overridden in `RiskMangerV1`.

Note: There is no need to performed the check for auctions fully filled, as the check is done in the `Deposit` smart contract.


contracts/RiskManagerV1.sol
```diff
@@ -214,6 +214,24 @@ contract RiskManagerV1 is Auctioneer, Ownable {
...
     }
 
+    /// @notice Throws if deposit not present among deposits in liquidation or
+    // the state of the deposit is not DEPOSIT_LIQUIDATION_IN_PROGRESS_STATE.
+    /// @dev This function is called when the auctioneer is informed about the
+    /// results of an auction. It is invoked from the Auctioneer contract's
+    /// offerTaken function.
+    /// @param auction Address of an auction whose deposit needs to be checked.
+    function checkDepositState(address auction) internal override {
+        require(
```
dimpar, line 224: I'm not sure if this check is needed here.  `checkDepositState()` can be called only internally through `takeOffer()` on an _opened_ auction. If an auction is closed (deleted by `harikari()`), we won't be able to call `takeOffer()` in the first place. I think it is enough to check the `deposit.currentState()` in case a deposit was bought outside the coverage pool.

test/liquidation/liquidation.test.js
```diff
@@ -95,4 +88,31 @@ describe("Integration -- liquidation happy path", () => {
       )
     })
   })
+
+  describe("when deposit has changed state between auction creation and auction offer taken", () => {
```
dimpar, line 92: I would add in the description "outside the coverage pool".  Maybe something like:
"when deposit has changed the state outside the coverage pool during pending coverage pool auction"

test/liquidation/liquidation.test.js
```diff
@@ -95,4 +88,31 @@ describe("Integration -- liquidation happy path", () => {
...
+    let auction
+    beforeEach(async () => {
+      auction = await prepareAuction()
+      // simulate deposit state change outside Coverge Pools
```
dimpar, line 96: typo: Coverge -> Coverage

contracts/Auctioneer.sol
```diff
@@ -146,4 +149,11 @@ contract Auctioneer is CloneFactory {
     ///         already closed and funds from the coverage pool are seized.
     /// @dev Override this function to act on auction closed as fully filled.
     function onAuctionFullyFilled(Auction auction) internal virtual {}
+
+    /// @notice Checks the state of the deposit asocciated with an auction
```
dimpar, line 153: Can we please change formatting? It seems that all other comments are aligned a bit differently, ie. new lines start directly beneath the first sentence, not under the <@tag> .

test/liquidation/liquidation.test.js
tomaszslabon: Fixed in https://github.com/keep-network/coverage-pools/pull/55/commits/e3816ce69b354db0468b9327952b4446a90885bc

test/liquidation/liquidation.test.js
tomaszslabon: Fixed in https://github.com/keep-network/coverage-pools/pull/55/commits/e3816ce69b354db0468b9327952b4446a90885bc

contracts/Auctioneer.sol
tomaszslabon: Sure, Fixed in d5e8369af46610a4d5129dde21d8d0fa85fa7998.

contracts/RiskManagerV1.sol
tomaszslabon: I added this check just to be on the safe side.
But, indeed, it is unnecessary (we cannot enter this function for a close auction).
Removed in  50997b5081f5d421e73bc66555abb64f14dbd672.

contracts/Auctioneer.sol
```diff
@@ -62,6 +62,9 @@ contract Auctioneer is CloneFactory {
     ) external {
         require(openAuctions[msg.sender], "Sender isn't an auction");
 
+        // Check the state of the deposit associated with the auction
+        checkDepositState(msg.sender);
```
pdyraga, line 66: `Auctioneer` should be abstract enough so it can be used by any risk manager implementation, also if that implementation does not have a concept of a deposit. I think we need function that is more generic, something that does not imply this contract works with tBTC.

Another note is that for `offerTaken` calls for fully filled auctions, [we already have the desired behavior](https://github.com/keep-network/tbtc/blob/762b189bdb1ce7b453aa5a4cd710e2092fdf378a/solidity/contracts/deposit/DepositLiquidation.sol#L140).

If instead of `checkDepositState` we do `onAuctionPartiallyFilled` next to `onAuctionFullyFilled`, we can have a more generic API and we can execute the deposit state check only for the latter given that it is already happening implicitly for the latter.

pdyraga: I'd also feel safer if we had those two scenarios covered later in a system test with real tBTC contract. But #53 needs to be merged first.

contracts/RiskManagerV1.sol
```diff
@@ -214,6 +214,20 @@ contract RiskManagerV1 is Auctioneer, Ownable {
...
 
+    /// @notice Throws if the state of the deposit is not
+    ///         DEPOSIT_LIQUIDATION_IN_PROGRESS_STATE.
+    /// @dev This function is called when the auctioneer is informed about the
+    ///      results of an auction. It is invoked from the Auctioneer
+    ///      contract's offerTaken function.
+    /// @param auction Address of an auction whose deposit needs to be checked.
+    function checkDepositState(address auction) internal override {
+        IDeposit deposit = IDeposit(auctionToDeposit[auction]);
+        require(
```
pdyraga, line 225: Worth dropping a comment on why do we need to execute this check (the deposit could be liquidated in the meantime, outside of coverage pools).

contracts/Auctioneer.sol
tomaszslabon: > If instead of checkDepositState we do onAuctionPartiallyFilled next to onAuctionFullyFilled, we can have a more generic API and we can execute the deposit state check only for the latter given that it is already happening implicitly for the latter.

I added function `onAuctionPartiallyFilled` and overridden it in `RiskManagerV1`:
```
function onAuctionPartiallyFilled(Auction auction) internal view override {
    IDeposit deposit = IDeposit(auctionToDeposit[address(auction)]);
    // Make sure the deposit was not liquidated outside of Coverage Pool
    require(
        deposit.currentState() == DEPOSIT_LIQUIDATION_IN_PROGRESS_STATE,
        "Deposit liquidation is not in progress"
    );
}
```

And the condition at the end of `offerTaken` looks like this:
```        
if (auction.isOpen()) {
    onAuctionPartiallyFilled(auction);
} else {
    onAuctionFullyFilled(auction);

    emit AuctionClosed(msg.sender);
    delete openAuctions[msg.sender];
}
```

tomaszslabon: > I'd also feel safer if we had those two scenarios covered later in a system test with real tBTC contract. But #53 needs to be merged first.

But should it be done in a separate PR?
Should we add a task to the backlog?

contracts/RiskManagerV1.sol
```diff
@@ -214,6 +217,20 @@ contract RiskManagerV1 is Auctioneer, Ownable {
         signerBondsSwapStrategy.swapSignerBonds{value: withdrawableAmount}();
     }
 
+    /// @notice Throws if the state of the deposit is not
+    ///         DEPOSIT_LIQUIDATION_IN_PROGRESS_STATE.
```
pdyraga, line 221: It does not *throw*, it *reverts* :wink: Also, let's explain why this could happen:
```
    /// @notice Reverts if the deposit for which the auction was created is no
    ///         longer in the liquidation state. This could happen if signer
    ///         bonds were purchased from tBTC deposit directly, outside of
    ///         coverage pool auction.
```

test/liquidation/liquidation.test.js
```diff
@@ -2,7 +2,7 @@ const { expect } = require("chai")
 const { to1e18 } = require("../helpers/contract-test-helpers")
```
pdyraga, line 2: This file is no longer only a liquidation happy path. Also, in #53, we introduced system tests under `test/system/liquidation.test.js`.

Given that we have more tests than just the liquidation happy path in both cases, I propose we do the rename to keep them as:
```
test/integration/integration.test.js
test/system/system.test.js
```
for now and if we define enough test cases that it will make sense, split those files into several smaller ones.

test/liquidation/liquidation.test.js
```diff
@@ -95,4 +88,31 @@ describe("Integration -- liquidation happy path", () => {
       )
     })
   })
+
+  describe("when deposit has changed the state outside the coverage pool during pending coverage pool auction", () => {
```
pdyraga, line 92: Maybe we can shorten it to just `when deposit has been liquidated by someone else`?

test/liquidation/liquidation.test.js
```diff
@@ -95,4 +88,31 @@ describe("Integration -- liquidation happy path", () => {
...
+    let auction
+    beforeEach(async () => {
+      auction = await prepareAuction()
+      // simulate deposit state change outside Coverage Pools
+      await tbtcDeposit.setStateLiquidated()
```
pdyraga, line 97: Instead of introducing a separate, test-only function, can we call `purchaseSignerBondsAtAuction()`? This is what actually happens in this case: third party purchasing signer bonds directly from tBTC deposit.

contracts/Auctioneer.sol
pdyraga: Yes, let's backlog it and work on it in a separate PR, no reason to block this one.

test/liquidation/liquidation.test.js
tomaszslabon: Fixed in 2c29566b2fba6d7c26e266b118d46d24f18e4d5c.

test/liquidation/liquidation.test.js
tomaszslabon: Fixed in 2c29566b2fba6d7c26e266b118d46d24f18e4d5c.

test/liquidation/liquidation.test.js
tomaszslabon: Done in 2c29566b2fba6d7c26e266b118d46d24f18e4d5c.

contracts/RiskManagerV1.sol
tomaszslabon: Fixed in 2c29566b2fba6d7c26e266b118d46d24f18e4d5c.

pdyraga (approved):

PR #56, opened 2021-05-21
Two step withdrawal period from the asset pool

pdyraga:
Withdrawal is now a two-step process with a withdrawal delay between
initiating and completing the withdrawal. It has been implemented to
avoid underwriters escaping from the pool just before the claim from the
risk manager (e.g. smart bots frontrunning risk manager/auction calls).

It takes 14 days at minimum for the underwriter to withdraw their
collateral and rewards. During that time, underwriter is still earning
rewards and their share of the pool is still a subject of a possible
coverage claim.

After 14 days withrawal delay, the underwriter has 7 days to complete
the withdrawal. After 7 days, graceful withdrawal timeout passes and
part of underwriter's rewards and collateral is slowly getting seized by
the pool - more and more over the time. This is to avoid smart
free-riders who wants to eat a cake and have a cake - be in the pool and
earn rewards but also being able to escape at any moment.

After additional 60 days (70 days from the time withdrawal delay
passed), 99% of tokens is seized by the pool and 1% of tokens is sent to
the notifier who completed the withdrawal on behalf of the underwriter.

```
     withdrawal              graceful withdrawal
       delay                 timeout
 /--------------\ /----------|
x----------------x-----------x-------------------------------x--------- time --------->
                 ^\------------------------------------------|
initiatedAt                                                   hard withdrawal
                                                              timeout
```

This two-step process does not work with the current
implementation of `EthAssetPool` given that the `EthAssetPool` becomes the
one initiating the withdrawal and for each next withdrawal, the previous
one gets overwritten in the `AssetPool` contract. The two-step withdrawal
needs much more complex accounting in `EthAssetPool`, probably with each
new ETH->WETH deposit swap being a separate contract. Given that we do not need
`EthAssetPool` for v1 release, there is no point in spending too much time
on reworking it right now. I deleted it and we can always take it back from git and
rework, once we really need it.


contracts/AssetPool.sol
```diff
@@ -10,6 +10,8 @@ import "@openzeppelin/contracts/access/Ownable.sol";
 import "./RewardsPool.sol";
 import "./UnderwriterToken.sol";
 
+import "hardhat/console.sol";
```
dimpar, line 13: oops

pdyraga: image

contracts/AssetPool.sol
```diff
@@ -64,11 +96,122 @@ contract AssetPool is Ownable {
...
+        );
+        /* solhint-disable not-rely-on-time */
+        withdrawalInitiatedTimestamp[msg.sender] = block.timestamp;
+
+        emit WithdrawalInitiated(
+            msg.sender,
+            pendingWithdrawal[msg.sender],
+            block.timestamp
+        );
+        /* solhint-enable not-rely-on-time */
```
dimpar, line 130: Do we need this guy?

pdyraga: I think yes, at least that's how we do it in all other places. Honestly, given how many times we need to ignore this detector, I am thinking about ignoring it in the global rules...

contracts/AssetPool.sol
```diff
@@ -64,11 +96,122 @@ contract AssetPool is Ownable {
...
+            "Underwriter token amount must be greater than 0"
+        );
+
+        underwriterToken.safeTransferFrom(msg.sender, address(this), covAmount);
+
+        pendingWithdrawal[msg.sender] = covAmount.add(
+            pendingWithdrawal[msg.sender]
+        );
+        /* solhint-disable not-rely-on-time */
+        withdrawalInitiatedTimestamp[msg.sender] = block.timestamp;
```
dimpar, line 131: just to make sure... If I'm an underwriter and I initiate my first withdrawal today for amount of 1000 and a week after I initiate a second withdrawal for amount of 1, then technically I will be able to withdraw my tokens 2 weeks after the second initiation (3 weeks after the first initiation), right? Basically, the latest initiation overrides whatever was initiated earlier?

pdyraga: Correct.

contracts/AssetPool.sol
```diff
@@ -80,17 +124,112 @@ contract AssetPool is Ownable {
...
         uint256 amountToWithdraw =
             covAmount.mul(collateralBalance).div(covSupply);
 
-        underwriterToken.safeTransferFrom(msg.sender, address(this), covAmount);
+        //
+        //      withdrawal              graceful withdrawal
+        //        delay                 timeout
+        //  /--------------\ /----------|
+        // x----------------x-----------x-------------------------------x------>
+        // ^                 \-----------------------------------------|
```
dimpar, line 181: it looks like the hard withdrawal timeout starts counting when withdrawal delay ends, but in the comment for the hard withdrawal it says that the time starts ticking at the moment of initiation. Can we maybe start the line from `initiatedAt` moment for the `hard withdrawal timeout`?

pdyraga: It starts counting after the withdrawal delay has passed, that comment is wrong, I will fix it.

contracts/AssetPool.sol
```diff
@@ -80,17 +124,112 @@ contract AssetPool is Ownable {
...
+            );
+            collateralToken.safeTransfer(underwriter, amountToWithdraw);
+        } else if (hardWithdrawalEndTimestamp >= block.timestamp) {
+            // After the graceful withdrawal timeout but before the hard
+            // withdrawal timeout. A portion of collateral and tokens is
+            // seized by the pool, proportionally to the time passed after
+            // the graceful withdrawal timeout.
+            uint256 delayRatio =
+                hardWithdrawalEndTimestamp.sub(block.timestamp).mul(1e18).div(
+                    hardWithdrawalTimeout.sub(gracefulWitdrawalTimeout)
```
dimpar, line 212: `hardWithdrawalTimeout.sub(gracefulWitdrawalTimeout)` seems this is a constant as well. Do you think we can make it a constant at the contract level?

pdyraga: We can, true. Though I am not sure if it's worth it, the gas cost is minimal and this code should ideally never be called - all underwriters are supposed to complete the withdrawal in the first 7 days. I do not even have an idea how to best name this const without complicating the code, heh.

contracts/AssetPool.sol
```diff
@@ -80,17 +124,112 @@ contract AssetPool is Ownable {
...
+                block.timestamp
+            );
+            collateralToken.safeTransfer(underwriter, amountToWithdraw);
+        } else if (hardWithdrawalEndTimestamp >= block.timestamp) {
+            // After the graceful withdrawal timeout but before the hard
+            // withdrawal timeout. A portion of collateral and tokens is
+            // seized by the pool, proportionally to the time passed after
+            // the graceful withdrawal timeout.
+            uint256 delayRatio =
+                hardWithdrawalEndTimestamp.sub(block.timestamp).mul(1e18).div(
```
dimpar, line 211: unlikely but we can have a corner case when `hardWithdrawalEndTimestamp == block.timestamp` and delayRatio = 0. Other transfers and calculations except emitting events won't make sense.

pdyraga: Ooof, good catch! Fixed.

contracts/AssetPool.sol
```diff
@@ -80,17 +124,112 @@ contract AssetPool is Ownable {
...
+                block.timestamp
+            );
+            emit GracefulWithdrawalTimedOut(underwriter, block.timestamp);
+            collateralToken.safeTransfer(underwriter, amountToWithdrawReduced);
+        } else {
+            // After the hard withdrawal timeout passed. 99% of tokens is seized
+            // by the pool, 1% of tokens goes go the notifier.
+            emit WithdrawalCompleted(underwriter, 0, block.timestamp);
+            emit WithdrawalTimedOut(underwriter, block.timestamp);
+            collateralToken.safeTransfer(msg.sender, amountToWithdraw.div(100));
```
dimpar, line 229: `amountToWithdraw.div(100)` I take that we operate on 1e18 numbers here and losing precision here is ok?

pdyraga: In this particular case, I do not think there is much we can improve. We could, theoretically, improve the precision of the previous case by computing `amountToWithdraw` inside the `else if` and not dividing by `covSupply` before calculating `delayRatio` and I tried to refactor the code this way but... it is getting really, really clumsy to understand and two other detectors start complaining that the code looks very similar in if/else blocks... Given that the vast majority (if not all) underwriters should complete the withdrawal in the first 7 days, I decided to pick the clarify of the code at the expense of precision. 🤷

dimpar (dismissed):
I looked only at the contract code.

contracts/AssetPool.sol
```diff
@@ -27,6 +27,46 @@ contract AssetPool is Ownable {
...
+    event WithdrawalCompleted(
+        address indexed underwriter,
+        uint256 amount,
+        uint256 timestamp
+    );
+    event GracefulWithdrawalTimedOut(
+        address indexed underwriter,
+        uint256 timestamp
+    );
+    event WithdrawalTimedOut(address indexed underwriter, uint256 timestamp);
```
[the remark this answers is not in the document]
pdyraga: That's something I was going about back and forward. We do, for example, add a timestamp in `AuctionLengthUpdateStarted` and all similar tBTC events also contain timestamps. I think the gas cost is minimal but having all the information needed inside the event could be helpful when crunching data for the dApp.

contracts/AssetPool.sol
```diff
@@ -80,17 +124,112 @@ contract AssetPool is Ownable {
...
+        uint256 withdrawalDelayEndTimestamp = initiatedAt.add(withdrawalDelay);
+        require(
+            withdrawalDelayEndTimestamp < block.timestamp,
+            "Withdrawal delay has not elapsed"
+        );
+
+        // slither-disable-next-line reentrancy-events
         rewardsPool.withdraw();
 
+        uint256 covAmount = pendingWithdrawal[underwriter];
```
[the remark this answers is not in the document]
pdyraga: Yes, we should! Done in https://github.com/keep-network/coverage-pools/pull/56/commits/f47236ecd2ec76b69142c2d8de90b81a2c3012e2.

lukasz-zimnoch (approved):

Issue #60, opened 2021-05-24
Withdrawal that reached the timeout can be reset 


montyly:
Severity: Low
Difficulty: Low

## Description
The hard withdrawal timeout can be prevented by creating a new withdrawal. As a result, users can avoid losing their assets due to inactivity.

Once the withdrawal delay is reached, there are three time windows:
https://github.com/keep-network/coverage-pools/blob/4aee098edbd4755d57dcfc7649ca80ab105500bb/contracts/AssetPool.sol#L179-L186

- The graceful withdrawal: the user receives all the assets
- After the graceful withdrawal: the user receives part of the assets, and the remainder is kept by the pool
- The hard withdrawal timeout: the caller receives 1% of the amount, and the pool the remainder

A user can avoid losing money due to the graceful withdrawal/hard withdrawal by call `inititalWithdrawal` again:
https://github.com/keep-network/coverage-pools/blob/4aee098edbd4755d57dcfc7649ca80ab105500bb/contracts/AssetPool.sol#L116-L131

Here `withdrawalInitiatedTimestamp` will be reset to the current timestamp, allowing the user to receive all the assets after the withdrawal delay.

## Exploit Scenario
Eve starts a withdrawal. Eve forgot to complete the withdrawal, and the withdrawal reaches the graceful withdrawal. Eve makes a new withdrawal of 1 wei. As a result, she avoids losing money.

## Recommendations
Short term, either:
- Prevent initiating a new withdrawal if the graceful withdrawal period is reached
- Document this behavior if it is the expected one

Long term, document the state machine representation of the withdrawal and ensure all the potential paths are explained and tested.

pdyraga:
Very good catch. Another attack vector here is that for the graceful withdrawal period, Eve is free-riding the pool. She can withdraw the declared funds by front-running claims. She can also minimize the withdrawal if the risk decreased as a result of some market movement.

I think the simplest strategy is the best here and leaves no room for manipulation: Eve can `initiateWithdrawal` one more time but only during the withdrawal delay. This action will reset the withdrawal clock. Each `initiateWithdrawal` call increases the amount Eve will eventually withdraw from the pool, i.e. initiated withdrawals sum up.

This way, Eve can decide to reduce her position in the pool even more without having to wait for the remaining time of the withdrawal delay to complete the pending withdrawal and then, waiting for the entire withdrawal delay again. At the same time, we do not leave room for manipulation and we keep the mechanism simple.

PR #63, opened 2021-05-25
Fix for audit issue "Withdrawal that reached the timeout can be reset"

tomaszslabon:
Closes: #60 

This pull requests addresses an issue reported during the audit: [Withdrawal that reached the timeout can be reset](https://github.com/keep-network/coverage-pools/issues/60).

Eve can `initiateWithdrawal` one more time but only during the withdrawal delay. This action will reset the withdrawal clock. Each `initiateWithdrawal` call increases the amount Eve will eventually withdraw from the pool, i.e. initiated withdrawals sum up.

This way, Eve can decide to reduce her position in the pool even more without having to wait for the remaining time of the withdrawal delay to complete the pending withdrawal and then, waiting for the entire withdrawal delay again. At the same time, we do not leave room for manipulation and we keep the mechanism simple.

### DIFF
--- a/contracts/AssetPool.sol
+++ b/contracts/AssetPool.sol
@@ -50,6 +50,16 @@ contract AssetPool is Ownable {
     // withdrawal on behalf of the underwriter. Hard withdrawal timeout starts
     // counting from the moment withdrawal delay has passed.
     uint256 public constant hardWithdrawalTimeout = 70 days;
+    // After the withdrawl initiation prevention timeout passes, the underwriter
+    // will not be able to initiate withdrawal again until the current
+    // withdrawal is completed. Before the withdrawal initiation prevention
+    // timeout passes, the underwriter can initiate withdrawal multiple times,
+    // and thus reset the withdrawal initiated timestamp to the timestamp of
+    // the last withdrawal initiation.
+    // The value of withdrawal initiation prevention timeout is the sum of the
+    // withdrawal delay and graceful withdrawal timeout.
+    uint256 public constant withdrawalInitiationPreventionTimeout =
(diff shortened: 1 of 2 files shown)


contracts/AssetPool.sol
```diff
@@ -123,10 +139,25 @@ contract AssetPool is Ownable {
...
             "Underwriter token amount must be greater than 0"
         );
+        // Ensure withdrawal not initiated or withdrawal initiation prevention
+        // timeout has not elapsed
+        //slither-disable-next-line incorrect-equality
+        require(
+            withdrawalInitiatedTimestamp[msg.sender] == 0 ||
+                withdrawalInitiatedTimestamp[msg.sender].add(
+                    withdrawalInitiationPreventionTimeout
+                ) >
```
pdyraga, line 149: If we do it this way, underwriters will be able to free-ride for some part of the graceful withdrawal period:
- Eve initiates the withdrawal for 100 000 KEEP.
- Eve waits for the withdrawal delay, earns rewards for that period.
- Eve waits for some part of the graceful withdrawal period.
- Now, Eve can withdraw immediately and even front-run coverage claims. She can also decide not to withdraw by initiating the withdrawal for 1e-18 KEEP if the risk lowered as a result of some market movement. 

It is being discussed [here](https://github.com/keep-network/coverage-pools/issues/60#issuecomment-848695084) and I would rather let her increase her withdrawal but only during the withdrawal delay period.

tomaszslabon: That makes sense.

pdyraga (approved):

pdyraga:
@tomaszslabon I let myself modify the PR description so that it reflects the recent code changes in this PR.

PR #66, opened 2021-05-26
Signer bonds Uniswap v2 strategy

lukasz-zimnoch:
Depends on: #53 (system tests are required for this pull request)

This pull request introduces the implementation of the `SignerBondsUniswapV2` contract. This contract provides an implementation of the `ISignerBondsSwapStrategy` interface and allows the following flow to take place:
1. Risk manager purchases signer bonds and deposits them on the `SignerBondsUniswapV2` contract
2. An interested third party (or the governance itself) swaps an arbitrary amount of bonds (within governance parameters boundaries) on UniswapV2 and sends acquired collateral tokens to the asset pool.
3. Acquired tokens increase the asset pool collateral balance without minting any underwriter tokens.

Several governance parameters can be set by the contract owner to keep swap transactions reasonable. These parameters are:
- `maxAllowedPriceImpact`: determines the maximum allowed price impact for swap transactions. Price impact depends on the transaction volume and acquired token reserve. Transactions whose impact exceeds the limit are reverted. This parameter forces transaction volumes to be reasonable and helps to avoid poor execution prices which are unacceptable from the business point of view.
- `slippageTolerance`: determines the slippage level which is tolerated by swap transactions. Slippage may occur in the time window between the moment when the swap transaction is submitted and eventually mined. Transactions whose slippage level exceeds the tolerance limit are reverted. 
- `swapDeadline`: determines the validity time of a swap transaction. If the transaction is not mined before the deadline it is reverted. 

This pull request also contains unit and system tests of the provided strategy. It also provides a test case that asserts that an arbitrary collateral token transfer doesn't break asset pool wthdrawals.


contracts/SignerBondsUniswapV2.sol
```diff
@@ -0,0 +1,254 @@
...
+    }
+
+    /// @notice Checks the price impact of buying a given amount of tokens
+    ///         against the maximum allowed price impact limit.
+    /// @param amount Amount of tokens.
+    /// @return True if the price impact is allowed, false otherwise.
+    function isAllowedPriceImpact(uint256 amount) internal view returns (bool) {
+        // Get reserve of the target token.
+        address WETH = uniswapRouter.WETH();
+        address token0 = WETH < targetToken ? WETH : targetToken;
```
dimpar, line 203: I'm a bit confused why comparing addresses will tell us which one is which. We can also add `token0()` to `IUniswapV2Pair` interface and get the address of `token0`. Then  `uint256 targetTokenReserve = WETH == token0 ? reserve1 : reserve0;` remains as is

lukasz-zimnoch: This code is based on https://github.com/Uniswap/uniswap-v2-periphery/blob/master/contracts/libraries/UniswapV2Library.sol. I think that comparing addressed using `<` is cheaper than calling a contract method.

contracts/SignerBondsUniswapV2.sol
```diff
@@ -0,0 +1,254 @@
...
+    event UniswapV2SwapExecuted(uint256[] amounts);
+
+    constructor(IUniswapV2Router _uniswapRouter, CoveragePool _coveragePool) {
+        uniswapRouter = _uniswapRouter;
+        targetTokenRecipient = address(_coveragePool.assetPool());
+        targetToken = address(_coveragePool.collateralToken());
+        uniswapPair = IUniswapV2Pair(
+            computePairAddress(
+                _uniswapRouter.factory(),
+                _uniswapRouter.WETH(),
```
dimpar, line 88: We use `uniswapRouter.WETH()` in a couple of places in this contract and the address of `WETH` is constant. Do you think we can assign `address WETH = uniswapRouter.WETH()` to a contract level and then reuse it?

contracts/SignerBondsUniswapV2.sol
```diff
@@ -0,0 +1,254 @@
...
+
+        return priceImpact <= priceImpactLimit;
+    }
+
+    /// @notice Compute Uniswap v2 pair address.
+    /// @param factory Address of the Uniswap v2 factory.
+    /// @param tokenA Address of token A.
+    /// @param tokenB Address of token B.
+    /// @return Address of token pair.
+    function computePairAddress(
```
dimpar, line 232: Can we call [getPair](https://uniswap.org/docs/v2/smart-contracts/factory/#getpair) instead which is a view function? We would need to add `IUniswapV2Factory`, but It should save us some gas by not calculating the pair address in this contract.

contracts/SignerBondsUniswapV2.sol
dimpar: Ah, I see now, when they [create](https://github.com/Uniswap/uniswap-v2-core/blob/master/contracts/UniswapV2Factory.sol#L23) a pair the tokens are sorted. Okay, I didn't realize that. Sure, let's leave it then.

contracts/SignerBondsUniswapV2.sol
lukasz-zimnoch: `getPair` will also consume gas because it will be called from within the contract creation transaction. It doesn't matter it's a view function. I have done some experiments using `gasleft()` and system tests. Here are the results:

The current approach:
```
computePairAddress(
    _uniswapRouter.factory(),
    _uniswapRouter.WETH(),
    targetToken
)
```
costs `7838` of gas.

The alternative:
```
IUniswapV2Factory(_uniswapRouter.factory()).getPair(_uniswapRouter.WETH(), targetToken);
```
costs `12825`.

dimpar: ok, I just looked at what gas reporter says, and indeed the cost of deployment in my implementation current vs alternative is `1605081` vs `1657021` but the function `swapSignerBondsOnUniswapV2()` diff is very little `155187` vs `155209`. Third parties like governance are interested in `swapSignerBondsOnUniswapV2()`. So I guess it boils down to code readability in this case and in large both versions are fine to me. What bugs me a tiny bit is just we copied some code from the 3rd party that can be accessed from their API. If the current approach looks fine to other devs, sure, let's leave it as is :)

contracts/SignerBondsUniswapV2.sol
lukasz-zimnoch: Yep, I thought about this at the beginning. However, the gas reporter says this is a more expensive alternative. Contract construction will cost roughly `6500` more. Each `swapSignerBondsOnUniswapV2` will cost around `500-750` more gas, depending on the final implementation.

contracts/SignerBondsUniswapV2.sol
lukasz-zimnoch: Such a change shouldn't affect `swapSignerBondsOnUniswapV2`. It uses the `uniswapPair` in the same way, no matter how it was obtained in the constructor. Regarding "copied some code from 3rd party" argument, please take a look at this doc: https://uniswap.org/docs/v2/smart-contract-integration/getting-pair-addresses. It's just one of the two ways to obtain a pair address. In this case, we're just calculating it leveraging the fact the pair contract was created using the [`CREATE2`](https://eips.ethereum.org/EIPS/eip-1014) opcode.

contracts/SignerBondsUniswapV2.sol
dimpar: Hm.. I see, yea, contract level vars are more expensive even if we just hardcode the WETH address. I'm okay what we have now.

contracts/SignerBondsUniswapV2.sol
dimpar: Right, I'm fine with both, but the same doc as you pointed says that the "canonical" way is the first one - `getPair`.

dimpar:
Rest looks good to me. Nice example on how to use Uniswap in our project, might be useful for other issues.

contracts/SignerBondsUniswapV2.sol
```diff
@@ -0,0 +1,260 @@
...
+        require(
+            _slippageTolerance <= BASIS_POINTS_DIVISOR,
+            "Maximum value is 10000 basis points"
+        );
+        slippageTolerance = _slippageTolerance;
+    }
+
+    /// @notice Sets the deadline for a swap transaction.
+    /// @param _swapDeadline Swap deadline in seconds. Value of this parameter
+    ///        should be grater than 0. It should be chosen carefully as
```
dimpar, line 139: > Value of this parameter should be grater than 0

Do we want to put a check `require` for it?

test/SignerBondsUniswapV2.test.js
```diff
@@ -0,0 +1,251 @@
...
+    })
+  })
+
+  describe("swapSignerBondsOnUniswapV2", () => {
+    const ethReserves = 1000
+    const tokenReserves = 5000
+    const exchangeRate = 5 // because one can get 5 tokens for every 1 ETH
+
+    beforeEach(async () => {
+      await uniswapV2PairStub.setReserves(
```
dimpar, line 163: If you're fine with the other comment regarding mocking `uniswapV2Pair`, then this can be replaced by the mock.

contracts/test/UniswapV2RouterStub.sol
```diff
@@ -0,0 +1,69 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity <0.9.0;
+
+import "@openzeppelin/contracts/token/ERC20/ERC20.sol";
```
dimpar, line 5: do we need this import?

contracts/test/UniswapV2RouterStub.sol
```diff
@@ -0,0 +1,69 @@
...
+import "../SignerBondsUniswapV2.sol";
+
+contract UniswapV2RouterStub is IUniswapV2Router {
```
dimpar, line 8: I guess it's a broader discussion for everyone be on the same page and agree on our internal standards whether we "hardhat mock" or use the Stubs. I think we can leave it as-is, but since hardhat offers mocking options it seems that everything here could've been mocked in tests. Maybe we should use Stubs only in cases where mocking for some reasons is not possible?

test/SignerBondsUniswapV2.test.js
```diff
@@ -0,0 +1,251 @@
...
+
+    // SignerBondsUniswapV2 has to point to the deployed UniswapV2PairStub
+    // instance to make tests work. However, before setting the new value
+    // via the stub, assert the initial value is correct and the pair
+    // address computing logic works well.
+    expect(await signerBondsUniswapV2.uniswapPair()).to.be.equal(
+      // address of real KEEP/ETH LP pool
+      "0xE6f19dAb7d43317344282F803f8E8d240708174a"
+    )
+    await signerBondsUniswapV2.setUniswapPair(uniswapV2PairStub.address)
```
dimpar, line 60: What do you think about mocking `uniswapV2Pair` contract? We can do something like this:
```
    mockUniswapPair = await deployMockContract(governance, IUniswapV2Pair.abi)
    await mockUniswapPair.mock.getReserves.returns(42, 32, 111111) // <- return whatever you want
    await signerBondsUniswapV2.setUniswapPair(mockUniswapPair.address)
```

By mocking ^ we can get rid of `UniswapV2PairStub.sol`

contracts/SignerBondsUniswapV2.sol
lukasz-zimnoch: It should be greater but doesn't have to. If we allow `0` we can allow the government to actually block swaps for some time. If the government will want to resign from this privilege, we can add the `require`.

contracts/test/UniswapV2RouterStub.sol
lukasz-zimnoch: Yep, we could use a Waffle mock in that case but I see two reasons why we shouldn't. First, I'd like to check the values of arguments `swapExactETHForTokens` is invoked with. Those arguments are crucial and we can't check them using Waffle's `to.be.calledOnContract` matcher as `@nomiclabs/hardhat-waffle` plugin doesn't support it at the moment. Secondly, I think having `UniswapV2RouterStub` with some kind of real Uniswap contract imitation gives better readability as we have everything in one place. Understanding how the test is supposed to work, is easier in that case, comparing to the alternative situation where we use Waffle mock and have multiple `mock.<method>.returns` scattered among the codebase.

test/SignerBondsUniswapV2.test.js
lukasz-zimnoch: In contrary to https://github.com/keep-network/coverage-pools/pull/66#discussion_r644244640, I think replacing `UniswapV2PairStub` with Waffle mock is a good idea, as this stub is very simple and we do some boilerplate here. WIll change it according to your proposition.

contracts/test/CoveragePoolStub.sol
```diff
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: MIT
 pragma solidity <0.9.0;
 
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
```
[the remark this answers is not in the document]
lukasz-zimnoch: Done https://github.com/keep-network/coverage-pools/pull/66/commits/cd57620af198ce557ef8d935bd042ab64be035d1

contracts/test/UniswapV2RouterStub.sol
lukasz-zimnoch: Nope. Done in https://github.com/keep-network/coverage-pools/pull/66/commits/cd57620af198ce557ef8d935bd042ab64be035d1

test/SignerBondsUniswapV2.test.js
lukasz-zimnoch: Please see https://github.com/keep-network/coverage-pools/pull/66/commits/c828b33c022f6043dbd98f812e8a44f7e058b75c

test/SignerBondsUniswapV2.test.js
lukasz-zimnoch: https://github.com/keep-network/coverage-pools/pull/66/commits/c828b33c022f6043dbd98f812e8a44f7e058b75c

lukasz-zimnoch:
Everything addressed @dimpar

contracts/SignerBondsUniswapV2.sol
```diff
@@ -0,0 +1,260 @@
...
+contract SignerBondsUniswapV2 is ISignerBondsSwapStrategy, Ownable {
+    using SafeMath for uint256;
+
+    // One basis point is equivalent to 1/100th of a percent.
+    uint256 public constant BASIS_POINTS_DIVISOR = 10000;
+
+    IUniswapV2Router public uniswapRouter;
+    IUniswapV2Pair public uniswapPair;
+    address public targetTokenRecipient;
+    address public targetToken;
```
pdyraga, line 65: I think it is better to be explicit that the recipient is `assetPool` and the target token is the `collateralToken`. It would give better transparency about the intentions.

contracts/SignerBondsUniswapV2.sol
```diff
@@ -0,0 +1,260 @@
...
+            );
+
+        emit UniswapV2SwapExecuted(amounts);
+    }
+
+    /// @notice Checks the price impact of buying a given amount of tokens
+    ///         against the maximum allowed price impact limit.
+    /// @param amount Amount of tokens.
+    /// @return True if the price impact is allowed, false otherwise.
+    function isAllowedPriceImpact(uint256 amount) internal view returns (bool) {
```
pdyraga, line 206: Wouldn't it be better to keep it public so that one can do a kind of a pre-flight before calling `swapSignerBondsOnUniswapV2`?

contracts/SignerBondsUniswapV2.sol
```diff
@@ -0,0 +1,260 @@
...
+        // Include slippage tolerance into the minimum amount of output tokens.
+        amountOutMin = amountOutMin
+            .mul(BASIS_POINTS_DIVISOR.sub(slippageTolerance))
+            .div(BASIS_POINTS_DIVISOR);
+
+        // TODO: Do not send acquired tokens to the asset pool in case an
+        //       open auction exists. In that case, additional tokens will
+        //       cause the auction to offer more of the coverage pool value for
+        //       the same price. This is unreasonable in terms of economy.
+        //       Figure out how to address that problem.
```
pdyraga, line 187: Feels like another governable parameter perhaps?

contracts/SignerBondsUniswapV2.sol
```diff
@@ -0,0 +1,260 @@
...
+    /// @notice Swaps signer bonds on Uniswap v2 exchange.
+    /// @dev Swaps the given ETH amount for the target token using the Uniswap
+    ///      decentralized exchange. The maximum ETH amount is capped by the
+    ///      contract balance. Some governance parameters are applied on the
+    ///      transaction. The swap's price impact must fit within the
+    ///      maximum allowed price impact and the transaction is constrained
+    ///      with the slippage tolerance and deadline. Acquired target tokens
+    ///      are sent to the recipient address set during contract construction.
+    /// @param amount Amount to swap.
+    function swapSignerBondsOnUniswapV2(uint256 amount) external {
```
pdyraga, line 158: Did not have a chance to dig in and it is not clear from the docs: is this function safe for "concurrent" use? Say there is 100 ETH to swap, then I execute a couple of transactions, one after another, and they all take some time to process on the Uniswap side. When they are all finally processed, successfully or not, does the state stay consistent?

lukasz-zimnoch: What do you mean by saying `does the state stay consistent?`? Regarding the rest, this is why we have a slippage tolerance parameter. Suppose three transactions: A, B, C. If they are sent one after another in that order, each of them will cause a specific price change (which depends on their price impact). If for example, A and B caused such a change that exceeds the slippage tolerance of C, C will be reverted.

contracts/SignerBondsUniswapV2.sol
lukasz-zimnoch: Yep. I'll add  this in a follow-up PR.

contracts/SignerBondsUniswapV2.sol
lukasz-zimnoch: Sure, we can do that: https://github.com/keep-network/coverage-pools/pull/66/commits/dc445aca15cb9a90c7eef077035a19e996a3960e

contracts/SignerBondsUniswapV2.sol
lukasz-zimnoch: Why not. See https://github.com/keep-network/coverage-pools/pull/66/commits/1788cc3fd025d7fb6e0bb34e55376bc7e21c01de

contracts/SignerBondsUniswapV2.sol
pdyraga: I mean the state of this contract - the amount of ETH it holds and so on. Say that there are 5 swaps pending, there is some outstanding ETH on this contract, three swaps succeed, two swaps fail and ETH comes back to this contract. Haven't had a chance to dig in so not saying it does not work, just asking :)

lukasz-zimnoch: Taking funds from this contract and executing a swap on Uniswap happens in the same transaction (Uniswap's method is payable). If something bad occurs during the swap, the whole transaction reverts, and funds remain untouched. 

Regarding multiple ones, if we have the following scenario:
1. Contract balance is `100 ETH`
2. Transaction A which swaps `80 ETH` is pending
3. Transaction B which swaps `90 ETH` is sent after transaction A but mined before them
4. `90 ETH` should be swapped and transaction A should revert due to insufficient balance. The final contract balance should be `10 ETH`

If we have:
1. Contract balance is `100 ETH`
2. Transaction A which swaps `40 ETH` is pending
3. Transaction B which swaps `50 ETH` is sent after transaction A but mined before them
4. Transaction A reverts due to, for example, slippage tolerance violation
5. `50 ETH` should be swapped and `40 ETH` from A should be untouched. The final contract balance should be `50 ETH`

tomaszslabon (commented):
I've looked through the code and didn't find any issues.

test/system/uniswap-v2.test.js
```diff
@@ -0,0 +1,144 @@
...
+      await expect(tx)
+        .to.emit(signerBondsUniswapV2, "UniswapV2SwapExecuted")
+        .withArgs([
+          "10000000000000000000", // ETH -> WETH
+          "89259521320540407673397", // WETH -> KEEP
+        ])
+    })
+
+    it("should consume a reasonable amount of gas", async () => {
+      await expect(parseInt(tx.gasLimit)).to.be.lessThan(11472000)
```
pdyraga, line 138: :scream: Something is wrong, block gas limit is around 12M now. 

Assuming 50Gwei gas price, one would need to have ~0.57 ETH on their account to execute this transaction.

contracts/SignerBondsUniswapV2.sol
dimpar: Ok, can we maybe add this info, something like "should be greater equal than 0". By reading this sentence in a comment I though it's a strict `>`, not `>=`?

contracts/test/UniswapV2RouterStub.sol
dimpar: Sure, let's leave it as-is as the Stub provides already a lot of functionality. But just want to mention one thing just for the future reference.

> Those arguments are crucial and we can't check them using Waffle's to.be.calledOnContract

Right, but I think there's a trick for this. Functions that return a value can be mocked with exactly the arguments that you expect. If the arguments won't match, then your expected return value won't be returned. We have the event `emit UniswapV2SwapExecuted(amounts);` after the `swapExactETHForTokens(args)` and `amounts` can be checked there, meaning that args for  `swapExactETHForTokens(args)` should match what you've mocked. When we deal with the functions that won't return anything, then hopefully in the future we can rely on `to.be.calledOnContract` as you mentioned.

```
          mockUniswapRouter.mock.swapExactETHForTokens.wtihArgs(
                amountOutMin,
                path,
                assetPool,
                block.timestamp
            ).returns(your_amount)
```
 If the arguments that you pass won't match the expected arguments, then the `your_amount` won't be returned.

contracts/SignerBondsUniswapV2.sol
lukasz-zimnoch: Ok done.

test/system/uniswap-v2.test.js
lukasz-zimnoch: There is something wrong with the estimate provided by Hardhat. Please see https://github.com/keep-network/coverage-pools/pull/66/commits/c3fda9c131035ca3d446b0575d526a942b6b1d38.

dimpar (approved):
looks good!

PR #68, opened 2021-05-28
Drafting calculation of tbtc auction value in percent

dimpar:
WIP.. this PR is addressing the following comment: https://github.com/keep-network/coverage-pools/pull/53#issuecomment-849686389

More prod. code and tests will be added after we agree on the general [approach](e258a25).

We need to calculate the percentage value of tbtc auction in liquidation with signer bonds in exchange for TBTC. Signer bonds are seized for the deposits in liquidation and we cannot use `deposit.collateralizationPercentage()` because it will return `0`. The idea in this PR is to use `deposit.auctionValue()` instead and read the current market ETH/BTC ratio. Based on the current collateral percentage value we can decide if a coverage pool auction should be opened or not.

### DIFF
--- a/contracts/RiskManagerV1.sol
+++ b/contracts/RiskManagerV1.sol
@@ -26,7 +26,7 @@ interface IDeposit {
 
     function lotSizeTbtc() external view returns (uint256);
 
-    function collateralizationPercentage() external view returns (uint256);
+    function auctionValue() external view returns (uint256);
 
     function withdrawableAmount() external view returns (uint256);
 }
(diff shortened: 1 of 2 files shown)


contracts/RiskManagerV1.sol
```diff
@@ -128,14 +128,25 @@ contract RiskManagerV1 is Auctioneer, Ownable {
             "Deposit is not in liquidation state"
         );
 
-        require(
-            deposit.collateralizationPercentage() <= collateralizationThreshold,
-            "Deposit collateralization is above the threshold level"
-        );
+        // TODO: Read ETHTBTC price ratio from ex. Uniswap. As of May 28 it is 0.0686243
```
pdyraga, line 131: Uniswap will not show BTC/ETH ratio, we need some other oracle like e.g. Maker DAO's oracle used by tBTC. To fetch information from the oracle, the contract needs to be whitelisted and I think this needs to go through the governance voting so it is not that simple.

dimpar: Yes, that's why I put TBTC in a comment, not BTC. Can't we use TBTC/ETH ratio from Uniswap?

pdyraga: TBTC usually trades with a premium compared to BTC given the minting costs.

pdyraga:
Closing without merging in favor of the approach proposed in #75.

PR #73, opened 2021-05-31
Update one less mapping when opening an auction

tomaszslabon:
Work in progress

This pull request aims to limit the amount of storage taken in the `RiskManagerV1` contract for `auction`/`deposit` mappings by removing  `auctionToDeposit`.

It seems to decrease the gas usage:
- `notifyLiquidation`         317220 => 294963
- `notifyLiquidated`          62345  =>  59779
- `publicCreateAuction`   252425 => 230146
- `takeOffer`                      92689  =>  92780

### DIFF
--- a/contracts/Auctioneer.sol
+++ b/contracts/Auctioneer.sol
@@ -29,7 +29,11 @@ contract Auctioneer is CloneFactory {
     // Holds the address of the auction contract
     // which will be used as a master contract for cloning.
     address public masterAuction;
-    mapping(address => bool) public openAuctions;
+
+    // Maps auction to the item being on offer for that auction (e.g. a tBTC
+    // deposit). Must be filled by the risk manager after `createAuction` was
+    // called.
+    mapping(address => address) public openAuctions;
 
     CoveragePool public coveragePool;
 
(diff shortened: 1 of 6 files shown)


contracts/Auctioneer.sol
```diff
@@ -29,7 +29,11 @@ contract Auctioneer is CloneFactory {
     // Holds the address of the auction contract
     // which will be used as a master contract for cloning.
     address public masterAuction;
-    mapping(address => bool) public openAuctions;
+
+    // Maps auction to the item being on offer for that auction (e.g. a tBTC
+    // deposit). Must be filled by the risk manager after `createAuction` was
+    // called.
+    mapping(address => address) public openAuctions;
```
lukasz-zimnoch, line 36: Why don't we clean up this mapping on auction regular/early close? We used to delete the mapping from the former `auctionToDeposit` map.

lukasz-zimnoch:
Please note the changes introduced by https://github.com/keep-network/coverage-pools/pull/88 before merging.

pdyraga:
I'd hold on with this PR until we merge all other changes and compile a report of gas usage for https://github.com/keep-network/coverage-pools/issues/81. The game may no longer be worth it.

pdyraga:
I would suggest we leave the code as-is. We cannot know what the next implementation of the risk manager will be and the current gas usage does not look that scary, at least if we eliminate the cost of calls to tBTC deposits that would be the same here anyway. What do you think @tomaszslabon ?

tomaszslabon:
> I would suggest we leave the code as-is. We cannot know what the next implementation of the risk manager will be and the current gas usage does not look that scary, at least if we eliminate the cost of calls to tBTC deposits that would be the same here anyway. What do you think @tomaszslabon ?
@pdyraga 
I agree. We should close this PR without merging.

Issue #74, opened 2021-06-01
Upgradeability mechanism


pdyraga:
All coverage pool contracts are non-upgradeable. Underwriters should be able to migrate their positions to a new version of coverage pool by moving their collateral and rewards to a new asset pool approved first by the governance without suffering the withdrawal delay.

PR #76, opened 2021-06-01
Upgradeability to a new Asset Pool

dimpar:
Refs: #74

Governance can approve a new Asset Pool from the CoveragePool contract. Underwriter will be able to move their collateral tokens from the old asset pool to a new one. Old underwriter tokens will be burned and new ones will be calculated, minted and then transferred back to an underwriter in exchange for their collateral tokens.


contracts/AssetPool.sol
```diff
@@ -253,13 +260,75 @@ contract AssetPool is Ownable {
...
+            block.timestamp
+        );
+
+        // mappings cleanup since all the underwriter's funds are transferred to
+        // a new asset pool
+        delete withdrawalInitiatedTimestamp[msg.sender];
+        delete pendingWithdrawal[msg.sender];
+
+        underwriterToken.burn(covUnderwriterBalance);
+        newUnderwriterToken.mint(msg.sender, covUnderwriterBalance);
```
pdyraga, line 304: Our token implementation has only one owner that is allowed to mint it. We can rework the token to allow for multiple minters but from a design perspective, it feels better if there was only one minter for a token.

pdyraga, line 304: I do not think we can assume COV amount minted by the second pool should be the same. It could happen that there was a claim and the math may not hold in this case.

contracts/AssetPool.sol
```diff
@@ -253,13 +260,75 @@ contract AssetPool is Ownable {
...
+        emit WithdrawalToNewAssetPoolCompleted(
+            msg.sender,
+            amountToWithdraw,
+            block.timestamp
+        );
+
+        // mappings cleanup since all the underwriter's funds are transferred to
+        // a new asset pool
+        delete withdrawalInitiatedTimestamp[msg.sender];
+        delete pendingWithdrawal[msg.sender];
```
pdyraga, line 301: I am not sure about it.... If we just delete them, and there is a pending withdrawal, the underwriter will lose a part of their tokens.

dimpar: Wouldn't [this](https://github.com/keep-network/coverage-pools/blob/upgradeability-for-asset-pool/contracts/AssetPool.sol#L274) line do the trick? I take that we're interested in moving the entire balance of underwriter to a new Asset Pool. Can't we read the total balance no matter if it's pending or not and then transfer it?

contracts/AssetPool.sol
dimpar: Okay, I take that the problem will pop up with the old AssetPool being an owner of `newUnderwriterToken` and a new AssetPool won't be able to mint `newUnderwriterTokens`? If yes, wdyt about transferring an ownership to a new AssetPool right after we mint new tokens, ie. `newUnderwriterToken.transferOwnership(address(newAssetPool))`?

dimpar: Do you mean that a claim could happen a transaction early? If yes, then a couple of lines above in this code we have the following, which reads the latest balance of the collateral token.
```
rewardsPool.withdraw(); 
uint256 collateralBalance = collateralToken.balanceOf(address(this));
```

pdyraga:
Answering the three ([1](https://github.com/keep-network/coverage-pools/pull/76#discussion_r645367225), [2](https://github.com/keep-network/coverage-pools/pull/76#discussion_r645372062), [3](https://github.com/keep-network/coverage-pools/pull/76#discussion_r645355202)) comments at once:
- I think only the given `AssetPool` should be eligible to mint `UnderwriterToken` and we should always have 1<>1 relationship between the pool and the token.
- If someone is exiting from the pool, they need to be able to complete their withdrawal. We can optionally let them cancel the withdrawal and migrate to the new pool but this case will be so rare that it does not justify the complexity we'd have to introduce to the contract.
- COV amounts between pools might not match. It is up to the new pool to decide about the amount of COV that should be minted based on the collateral transferred and the current pool's balance.

Here are the building blocks I think we need (disclaimer: I haven't had a chance to validate it in the code):
- `approveNewPoolUpgrade(address newAssetPool)` for the governance
- `upgradeToNewPool(uint256 covAmount)` for the underwriter - needs to transfer COV from the underwriter, compute the collateral value and send the collateral to the new pool
- `deposit(uint256 amount, address underwriter)` - just like the standard deposit except that mints COV to `underwriter` and not `msg.sender`.

dimpar:
@pdyraga can you pls take a look again?

contracts/AssetPool.sol
```diff
@@ -17,16 +18,19 @@ import "./UnderwriterToken.sol";
...
 ///         Asset Pool.
-contract AssetPool is Ownable {
+contract AssetPool is Ownable, IAssetPool {
     using SafeERC20 for IERC20;
     using SafeERC20 for UnderwriterToken;
     using SafeMath for uint256;
 
     IERC20 public collateralToken;
     UnderwriterToken public underwriterToken;
+    UnderwriterToken public newUnderwriterToken;
```
pdyraga, line 28: We could have a `depositFor(uint256 amount, address covRecipient)` function. This would simplify the design - we wouldn't have to compute the balance of new COV when updating and we wouldn't need this reference.

contracts/AssetPool.sol
```diff
@@ -66,6 +70,12 @@ contract AssetPool is Ownable {
         uint256 timestamp
     );
     event WithdrawalTimedOut(address indexed underwriter, uint256 timestamp);
+    event UpgradeToNewAssetPoolCompleted(
```
pdyraga, line 73: Maybe just `AssetPoolUpgraded`? This is just a one-step process, so there is no initiation & completion.

contracts/AssetPool.sol
```diff
@@ -66,6 +70,12 @@ contract AssetPool is Ownable {
...
+        address indexed underwriter,
+        uint256 collateralAmount,
+        uint256 covAmount,
```
pdyraga, line 76: I think this should be the COV amount from this pool. This is what `WithdrawalInitiated` includes and this is the event from the current contract, so should inform about the current contract's state changes.

contracts/AssetPool.sol
```diff
@@ -253,6 +263,94 @@ contract AssetPool is Ownable {
...
+            address(_newAssetPool) != address(0),
+            "New asset pool can't be zero address"
+        );
+        require(
+            address(_newUnderwriterToken) != address(0),
+            "New underwriter token can't be zero address"
+        );
+
+        newAssetPool = _newAssetPool;
+        newUnderwriterToken = _newUnderwriterToken;
```
pdyraga, line 351: Worth emitting an event.

contracts/AssetPool.sol
```diff
@@ -253,6 +263,94 @@ contract AssetPool is Ownable {
...
+        underwriterToken.burnFrom(msg.sender, covAmount);
+    }
+
+    /// @notice Allows governance to set a new asset pool and a new underwriter
+    ///         token so the underwriters can move their collateral tokens to
+    ///         a new asset pool.
+    function approveNewAssetPoolUpgrade(
+        IAssetPool _newAssetPool,
+        UnderwriterToken _newUnderwriterToken
+    ) external onlyOwner {
```
pdyraga, line 340: The owner of this contract is the coverage pool so this will not work unless the idea was to route this call through `CoveragePool` which I think is acceptable and probably the best solution here.

contracts/interfaces/IAssetPool.sol
```diff
@@ -0,0 +1,11 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity <0.9.0;
+
+interface IAssetPool {
+    /// @notice Accepts the given amount of collateral token as a deposit and
+    ///         mints underwriter tokens representing pool's ownership.
+    /// @dev Before calling this function, collateral token needs to have the
+    ///      required amount accepted to transfer to the asset pool.
+    function deposit(uint256 amount) external;
```
pdyraga, line 10: I would also add `initiateWithdrawal` and `completeWithdrawal` to this interface. Withdrawal will always be a two-step process and even if not, only deposit function is needed for the upgrade.

contracts/AssetPool.sol
pdyraga: The link is pointing now to a solhint line so I am not sure what was there :) But I think we should keep upgrade and withdrawal separate for simplicity.

contracts/AssetPool.sol
pdyraga: I do not think this helps. New asset pool and old asset pool have their own COV supplies and collateral token balances.

contracts/AssetPool.sol
dimpar: I think that was the initial approach, but based on [this](https://www.flowdock.com/app/cardforcoin/tech/threads/kLsWXXDV8tcJ2uXtKBQqxqDAUPA) conversation, you mentioned we won't need to pass underwriter and that's why we need `UnderwriterToken` here to make it happen.

Okay, I'll change it to the initial approach.

contracts/AssetPool.sol
dimpar: The reason behind it was that this contract was transferring new underwriter tokens to underwriters, but since the latest approach is that a new asset pool will handle transfer of new cov tokens we can change it

contracts/AssetPool.sol
dimpar: Yup, that's the idea, will add this change in the next commits.

contracts/interfaces/IAssetPool.sol
dimpar: Done

contracts/AssetPool.sol
dimpar: Done

contracts/AssetPool.sol
dimpar: Done

contracts/AssetPool.sol
dimpar: yea, that comment was not relevant anymore.

contracts/AssetPool.sol
```diff
@@ -29,16 +31,19 @@ import "./UnderwriterToken.sol";
...
     using SafeERC20 for UnderwriterToken;
     using SafeMath for uint256;
 
     IERC20 public collateralToken;
     UnderwriterToken public underwriterToken;
+    UnderwriterToken public newUnderwriterToken;
 
     RewardsPool public rewardsPool;
 
+    IAssetPoolUpgrade public newAssetPool;
```
pdyraga, line 45: This is no longer needed, right?

dimpar: [here](https://github.com/keep-network/coverage-pools/blob/upgradeability-for-asset-pool/contracts/AssetPool.sol#L322) we call `depositFor()` on a new Asset Pool. Before we can call it, the governance has to [approve](https://github.com/keep-network/coverage-pools/blob/upgradeability-for-asset-pool/contracts/AssetPool.sol#L347) it. We need `newAssetPool` if we do it this way.

pdyraga: Oops I am sorry, GitHub tricked me, I probably did not refresh the page before posting this comment. This comment was for line 41 with `newUnderwritertToken`.

dimpar: Ah, right `newUnderwriterToken` is not needed. Removed f757e4622d79e20c4e7c1e07c1530d407fab334a

test/AssetPool.test.js
```diff
@@ -975,4 +975,243 @@ describe("AssetPool", () => {
...
+
+        beforeEach(async () => {
+          await assetPool
+            .connect(coveragePool)
+            .approveNewAssetPoolUpgrade(newAssetPool.address)
+
+          await assetPool.connect(underwriter1).deposit(amountToDeposit)
+          await underwriterToken
+            .connect(underwriter1)
+            .approve(assetPool.address, amountToDeposit)
```
pdyraga, line 1109: Shouldn't `amountToUpgrade` be approved instead?

test/AssetPool.test.js
```diff
@@ -975,4 +975,243 @@ describe("AssetPool", () => {
...
+          await assetPool
+            .connect(coveragePool)
+            .approveNewAssetPoolUpgrade(newAssetPool.address)
+          await assetPool.connect(underwriter1).deposit(amountToDeposit)
+          await assetPool
+            .connect(coveragePool)
+            .claim(coveragePool.address, claim)
+          await underwriterToken
+            .connect(underwriter1)
+            .approve(assetPool.address, amountToDeposit)
```
pdyraga, line 1159: Shouldn't `amountToUpgrade` be approved instead?

test/AssetPool.test.js
```diff
@@ -975,4 +975,243 @@ describe("AssetPool", () => {
...
+          await collateralToken
+            .connect(rewardManager)
+            .approve(rewardsPool.address, allocatedReward)
+          await rewardsPool.connect(rewardManager).topUpReward(allocatedReward)
+
+          await assetPool.connect(underwriter1).deposit(amountToDeposit)
+
+          await underwriterToken
+            .connect(underwriter1)
+            .approve(assetPool.address, amountToDeposit)
```
pdyraga, line 1198: Shouldn't `amountToUpgrade` be approved instead? Also, shouldn't we approve after one day passed. Should not do any difference technically but this is what will happen in the real world.

contracts/AssetPool.sol
```diff
@@ -266,6 +277,77 @@ contract AssetPool is Ownable {
...
     }
 
+    /// @notice Transfers collateral tokens to a new Asset Pool which previously
+    ///         was approved by the governance.
+    ///         Old underwriter tokens are burned in favor of new tokens minted
+    ///         in a new Asset Pool. New tokens are sent directly to the
+    ///         underwriter from a new Asset Pool.
+    /// @param covAmount Amount of underwriter tokens used to calculate collateral
+    ///                  tokens which are transferred to a new asset pool.
+    function upgradeToNewAssetPool(uint256 covAmount) external {
```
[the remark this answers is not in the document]
dimpar: e6dc71d

contracts/interfaces/IAssetPoolUpgrade.sol
```diff
@@ -0,0 +1,24 @@
...
+//
+//                           Trust math, not hardware.
+
+// SPDX-License-Identifier: MIT
+
+pragma solidity <0.9.0;
+
+import "./IAssetPool.sol";
+
+interface IAssetPoolUpgrade is IAssetPool {
```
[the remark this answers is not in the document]
dimpar: But at the same time we are interested in a new Asset Pool conforming to `IAssetPool` as the old one does. Devs will have to remember to add `IAssetPool` to `NewAssetPool is IAssetPoolUpgrade, IAssetPool` cause it won't be caught at compile time. 
Having 3 empty functions in a test stub is no biggy imo, just shows what functions a real contract has to have. But anyway, I've separated them per your comment if you feel they should be separated. a6f7056

test/AssetPool.test.js
```diff
@@ -975,4 +975,243 @@ describe("AssetPool", () => {
...
+      const NewUnderwriterToken = await ethers.getContractFactory(
+        "UnderwriterToken"
+      )
+      newUnderwriterToken = await NewUnderwriterToken.deploy(
+        "New Underwriter Token",
+        "newCOV"
+      )
+      await newUnderwriterToken.deployed()
+
+      const NewAssetPool = await ethers.getContractFactory("NewAssetPoolStub")
```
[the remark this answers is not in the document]
dimpar: 18fed51

test/AssetPool.test.js
```diff
@@ -975,4 +975,243 @@ describe("AssetPool", () => {
...
+      newAssetPool = await NewAssetPool.deploy(
+        collateralToken.address,
+        newUnderwriterToken.address
+      )
+      await newAssetPool.deployed()
+    })
+
+    context("when upgrading to a new asset pool as an owner", () => {
```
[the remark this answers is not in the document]
dimpar: 18fed51

test/AssetPool.test.js
```diff
@@ -975,4 +975,243 @@ describe("AssetPool", () => {
...
+
+    context("when upgrading with zero underwriter tokens", () => {
+      it("should revert", async () => {
+        await expect(
+          assetPool.connect(underwriter1).upgradeToNewAssetPool(0)
+        ).to.be.revertedWith("Underwriter token amount must be greater than 0")
+      })
+    })
+
+    context("when upgrading with amount less than available", () => {
```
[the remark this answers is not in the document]
dimpar: 18fed51

test/AssetPool.test.js
```diff
@@ -975,4 +975,243 @@ describe("AssetPool", () => {
...
+          expect(tx)
+            .to.emit(assetPool, "AssetPoolUpgraded")
+            .withArgs(
+              underwriter1.address,
+              amountToUpgrade,
+              amountToUpgrade,
+              await lastBlockTime()
+            )
+        })
+      })
```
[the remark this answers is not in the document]
dimpar: 18fed51

test/AssetPool.test.js
dimpar: 18fed51

test/AssetPool.test.js
dimpar: 18fed51

test/AssetPool.test.js
dimpar: 18fed51

contracts/CoveragePool.sol
```diff
@@ -149,4 +151,13 @@ contract CoveragePool is Ownable {
                 .mul(portionToSeize)
                 .div(CoveragePoolConstants.FLOATING_POINT_DIVISOR);
     }
+
+    /// @notice Approves upgradeability of a new asset pool.
+    /// @param _newAssetPool New asset pool
+    function approveNewAssetPoolUpgrade(IAssetPoolUpgrade _newAssetPool)
+        external
```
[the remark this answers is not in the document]
dimpar: Hm, not sure what was changed, but it's not complaining when I run it.

test/AssetPool.test.js
```diff
@@ -975,4 +975,278 @@ describe("AssetPool", () => {
...
+      newAssetPool = await NewAssetPool.deploy(
+        collateralToken.address,
+        newUnderwriterToken.address
+      )
+      await newAssetPool.deployed()
+      await newUnderwriterToken.transferOwnership(newAssetPool.address)
+    })
+
+    context(
+      "when a governance was compromised and it approved a fake new asset pool",
```
pdyraga, line 1057: Can we say instead "when new asset pool address does not match"? This could be a user error as well. Or, in extreme circumstances, upgrade of an upgrade conflicting with a pending underwriter's transaction.

pdyraga: Actually... we can leave it as is, after reading the test one more time I see what was the reasoning.

test/AssetPool.test.js
pdyraga: > Also, shouldn't we approve after one day passed. Should not do any difference technically but this is what will happen in the real world.

This still needs to be addressed. For:
- when there was a claim
- when rewards were allocated

test/AssetPool.test.js
dimpar: fcab985 Agree on the latter, but I took that the whole point for adding this check in contract was because something bad can happen with the governance, hence the described scenerio context.

test/AssetPool.test.js
dimpar: Done in 089f488

test/AssetPool.test.js
```diff
@@ -9,7 +9,7 @@ const {
 
 const RewardsPoolJSON = require("../artifacts/contracts/RewardsPool.sol/RewardsPool.json")
 
-describe("AssetPool", () => {
+describe.only("AssetPool", () => {
```
pdyraga, line 12: `only` 😬

dimpar: oops.. removed

test/AssetPool.test.js
```diff
@@ -194,6 +194,8 @@ describe("AssetPool", () => {
       const allocatedReward = to1e18(70)
 
       beforeEach(async () => {
+        await increaseTime(86400) // 1 day
```
pdyraga, line 197: This is not needed here, or is it?

dimpar: eh, not sure how it got there... 83048dd

pdyraga (approved):

PR #79, opened 2021-06-02
 Validate deposit's bond auction value

### LINKED ISSUE #75: Replace deposit collateralization validation
To minimize gas usage and claims on the coverage pool to the required minimum, tBTC v1 risk manager opens auctions only for liquidations no one else is willing to take. In practice, it is expected that arb bots will purchase signer bonds for most of the liquidated tBTC deposits before the risk manager will be required to open an auction.

The initial plan for implementing this feature was to have a governable collateralization threshold in the risk manager - the risk manager accepts to work with the deposit only if it is in a liquidation state and only if it has collateralization below the threshold value. This approach is problematic because tBTC `Deposit` seizes signer bonds when it is transitioning to liquidating state and for liquidating deposit `collateralizationPercentage()` function returns 0.

Instead of using collateralization threshold, we can check the auction value. If 100% [of the bond is on the auction](https://github.com/keep-network/tbtc/blob/master/solidity/contracts/deposit/DepositUtils.sol#L292) (make this % governable), the risk manager accepts to work with the deposit.

lukasz-zimnoch:
Closes #75 

The risk manager should not open an auction for a deposit if the deposit's bond auction offers a bond percentage lower than a `bondAuctionThreshold`.

This is because the risk manager should open a coverage pool auction for only those deposits that nobody else is willing to purchase. It is expected that arbitrage bots will purchase signer bonds for most of the liquidated tBTC deposits before the risk manager will be required to open an auction.

The `bondAuctionThreshold` parameter can be changed by the governance after preserving the governance delay.

This pull request also gets rid of the `collateralizationThreshold` which was the first approach of the aforementioned requirement.

### DIFF
--- a/contracts/RiskManagerV1.sol
+++ b/contracts/RiskManagerV1.sol
@@ -38,9 +38,9 @@ interface IDeposit {
 
     function lotSizeTbtc() external view returns (uint256);
 
-    function collateralizationPercentage() external view returns (uint256);
-
     function withdrawableAmount() external view returns (uint256);
+
+    function auctionValue() external view returns (uint256);
 }
 
 /// @title ISignerBondsSwapStrategy
@@ -61,16 +61,16 @@ contract RiskManagerV1 is Auctioneer, Ownable {
 
     uint256 public constant DEPOSIT_LIQUIDATION_IN_PROGRESS_STATE = 10;
     uint256 public constant DEPOSIT_LIQUIDATED_STATE = 11;
-    // Auction will not be opened if the deposit collateralization level does not
-    // fall below this collateralization threshold.
+    // Coverage pool auction will not be opened if the deposit's bond auction
+    // offers a bond percentage lower than this threshold.
     // Risk manager should open a coverage pool auction for only those deposits
     // that nobody else is willing to purchase. The default value can be updated
     // by the governance in two steps. First step is to begin the update process
     // with the new value and the second step is to finalize it after
     // GOVERNANCE_TIME_DELAY has passed.
-    uint256 public collateralizationThreshold; // percent
-    uint256 public newCollateralizationThreshold;
-    uint256 public collateralizationThresholdChangeInitiated;
+    uint256 public bondAuctionThreshold; // percent
+    uint256 public newBondAuctionThreshold;
+    uint256 public bondAuctionThresholdChangeInitiated;
 
     uint256 public auctionLength;
     uint256 public newAuctionLength;
@@ -95,11 +95,11 @@ contract RiskManagerV1 is Auctioneer, Ownable {
     event AuctionLengthUpdateStarted(uint256 auctionLength, uint256 timestamp);
     event AuctionLengthUpdated(uint256 auctionLength);
 
-    event CollateralizationThresholdUpdateStarted(
-        uint256 collateralizationThreshold,
+    event BondAuctionThresholdUpdateStarted(
+        uint256 bondAuctionThreshold,
         uint256 timestamp
     );
-    event CollateralizationThresholdUpdated(uint256 collateralizationThreshold);
+    event BondAuctionThresholdUpdated(uint256 bondAuctionThreshold);
 
     event SignerBondsSwapStrategyUpdateStarted(
         address indexed signerBondsSwapStrategy,
@@ -129,12 +129,12 @@ contract RiskManagerV1 is Auctioneer, Ownable {
         ISignerBondsSwapStrategy _signerBondsSwapStrategy,
         address _masterAuction,
         uint256 _auctionLength,
-        uint256 _collateralizationThreshold
+        uint256 _bondAuctionThreshold
     ) Auctioneer(_coveragePool, _masterAuction) {
         tbtcToken = _tbtcToken;
         signerBondsSwapStrategy = _signerBondsSwapStrategy;
         auctionLength = _auctionLength;
-        collateralizationThreshold = _collateralizationThreshold;
+        bondAuctionThreshold = _bondAuctionThreshold;
     }
 
     /// @notice Receive ETH from tBTC for purchasing & withdrawing signer bonds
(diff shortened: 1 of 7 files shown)


pdyraga:
@lukasz-zimnoch Could you please solve conflicts in this PR?

contracts/RiskManagerV1.sol
```diff
@@ -38,9 +38,9 @@ interface IDeposit {
```
pdyraga, line 38: Not in this PR but just noticed it - we could do something to remove `notifyRedemptionSignatureTimedOut` from `IDeposit` interface. This function is called only in tests and ideally, `IDeposit` should have only those functions the risk manager is interested in.

contracts/test/DepositStub.sol
```diff
@@ -77,4 +78,8 @@ contract DepositStub is IDeposit {
     function withdrawableAmount() external view override returns (uint256) {
         return address(this).balance;
     }
+
+    function setAuctionValue(uint256 _auctionValue) external {
+        auctionValue = _auctionValue;
```
pdyraga, line 83: Linter is complaining about the function order.

contracts/test/DepositStub.sol
```diff
@@ -77,4 +78,8 @@ contract DepositStub is IDeposit {
     function withdrawableAmount() external view override returns (uint256) {
         return address(this).balance;
     }
+
```
pdyraga, line 81: What do you think about separating this function from the rest for clarity what comes from tBTC and what was added by us? Something like:
```
    //
    // Not in tBTC deposit interface, functions below were added just for tests.
    //

    function setAuctionValue(uint256 _auctionValue) external {
        auctionValue = _auctionValue;
    }
```

test/system/liquidation.test.js
```diff
@@ -136,6 +138,22 @@ describeFn("System -- liquidation", () => {
...
       await tbtcDeposit.notifyRedemptionSignatureTimedOut()
+
+      // The deposit's auction must offer at least 75% of bonds to be accepted
+      // by the risk manager. At starting block, the deposit's auction exposes
+      // 66% so an immediate `notifyLiquidation` must revert.
+      await expect(
+        riskManagerV1.notifyLiquidation(tbtcDeposit.address)
+      ).to.revertedWith(
+        "Deposit bond auction percentage is below the threshold level"
+      )
```
pdyraga, line 149: This could be a separate test that the `notifyLiquidation` call is reverted. WDYT?

test/RiskManagerV1.test.js
```diff
@@ -331,31 +328,33 @@ describe("RiskManagerV1", () => {
...
         await tbtcToken.mint(bidder.address, auctionLotSize)
         await tbtcToken.connect(bidder).approve(auctionAddress, auctionLotSize)
         auction = new ethers.Contract(auctionAddress, Auction.abi, owner)
-        auction.connect(bidder).takeOffer(to1ePrecision(25, 16))
+        await auction.connect(bidder).takeOffer(to1ePrecision(25, 16))
 
         // Simulate that deposit was liquidated by someone else.
-        await mockIDeposit.mock.currentState.returns(depositLiquidatedState)
+        await tbtcToken.mint(owner.address, auctionLotSize)
+        await tbtcToken.approve(depositStub.address, auctionLotSize)
```
pdyraga, line 342: Can we do `tbtcToken.connect(owner).approve(...)` and `depositStub.connect(owner).purchaseSignerBondsAtAuction` for clarity? I was not sure at first why we are minting tokens to owner's account.

pdyraga (commented):
Four small notes, otherwise looks good.

test/system/liquidation.test.js
lukasz-zimnoch: It's not that easy as system tests should ideally use `before` not `beforeEach` which makes test initialization very hard and cumbersome.

contracts/RiskManagerV1.sol
lukasz-zimnoch: In one of the previous PRs I introduced `IDepositStub` interface exactly for that purpose. However, we decided to move all real `Deposit` methods to the `IDeposit` interface, even if they are not needed by contracts.

contracts/test/DepositStub.sol
lukasz-zimnoch: If we want to address https://github.com/keep-network/coverage-pools/pull/79#discussion_r644598509 it must be at the end. After all, it's just a warning 🤷

contracts/test/DepositStub.sol
lukasz-zimnoch: https://github.com/keep-network/coverage-pools/pull/79/commits/a7745dcdc76f981f1d1643f861db76b94abcc422

test/RiskManagerV1.test.js
lukasz-zimnoch: https://github.com/keep-network/coverage-pools/pull/79/commits/17844aa27212eb132b5743a7088459cca6fa4217

contracts/RiskManagerV1.sol
pdyraga: Hmm... I think we could have a misunderstanding. If I am not mistaken we were using `DepositStub` and we were additionally adding `IDepositStub`. Now we have a system test where we cannot use `DepositStub` (or at least, we shouldn't be).

Anyway, it does not matter now. 

If we introduce this interface in `contracts/test/DepositStub.sol`, document that those are the functions from tBTC Deposit but used only for tests, and even maybe consider naming this interface just `IDeposit` (?) this will be better than what we have now IMO. What do you think?

Anyway, no reason to block this PR, let's see what we can do about it separately.

contracts/test/DepositStub.sol
pdyraga: ... and it's a warning in a stub. Yeah :+1:

contracts/RiskManagerV1.sol
lukasz-zimnoch: Yep, we could try it as you proposed.

test/system/liquidation.test.js
pdyraga: I'll leave this thread open as a reminder, just in case we discover a better way one day.

pdyraga (approved):

PR #86, opened 2021-06-04
Risk manager can be quickly approved after deployment

tomaszslabon:
This pull request adds functionality that allows one risk manager (the first one after deployment) to be approved by `CoveragePool` without any governance delay.
Without it, it would be necessary to wait 30 days after approving the first risk manager before any `takeOffer` calls on any auction (risk manager has to be approved to `seizeFunds` from asset pool). 
All subsequent risk managers will have to wait for the governance delay to elapse.

**Suggestion: We could think about adding a function `getRemainingRiskManagerApprovalTime(address riskManager)` and return  the time remaining for its approval.**


pdyraga:
Instead of injecting this logic into finalization, what if we had a function that could be called only one time? We would be able to initialize the coverage pool in one transaction, the intention will be clearer, and `getRemainingRiskManagerApprovalTime` would not be cluttered with the first approval logic.

What I'd suggest:
- `approveFirstRiskManager` - can be called only one time if `!firstRiskManagerApproved`
- `beginRiskManagerApproval`, just like in main
- `finalizeRiskManagerApproval`, just like in main

Also, addressing the note from the PR description, I suggest we take the opportunity and do three additional changes in this PR:
- implement `getRemainingRiskManagerApprovalTime`
- add `RiskManagerApprovalStarted` event
- add `RiskManagerApprovalCompleted` event

tomaszslabon:
@pdyraga 
I think I addressed all your remarks: 
- added the `approveFirstRiskManager` function in 84a4f0b5cff12240156f861511eb180b4e3d57e2 
- added `getRemainingRiskManagerApprovalTime` function and even emitting in 716cc840caf2571168d0308b90e5b6df7d722897.

contracts/CoveragePool.sol
```diff
@@ -23,24 +23,29 @@ import "@openzeppelin/contracts/math/SafeMath.sol";
...
+
     // Currently approved risk managers
     mapping(address => bool) public approvedRiskManagers;
     // Timestamps of risk managers whose approvals have been initiated
     mapping(address => uint256) public riskManagerApprovalTimestamps;
     // Timestamps of risk managers whose unapprovals have been initiated
     mapping(address => uint256) public riskManagerUnapprovalTimestamps;
 
-    /// @notice Throws if called by a risk manager that has not been approved
+    event RiskManagerApprovalStarted(address indexed riskManager);
```
pdyraga, line 45: Can we add `uint256 timestamp` to `RiskManagerApprovalStarted` event? We use this pattern for governance-related events in risk manager. I'd also suggest to not index `riskManager` address on events. It is not needed in this case - those events will be emitted rarely and risk manager address should always be the only and unique value.

pdyraga: ... Since we are here can we add events for the unapproval process as well? 😬 Same about get remaining time function.

contracts/CoveragePool.sol
```diff
@@ -51,19 +56,33 @@ contract CoveragePool is Ownable {
...
     }
 
+    /// @notice Approves the first risk manager
+    /// @dev Can be called only by the contract owner. Can be called only once.
+    ///      Does not require any further calls to any functions.
+    /// @param riskManager Risk manager that will be approved.
+    function approveFirstRiskManager(address riskManager) external onlyOwner {
+        require(
+            !firstRiskManagerApproved,
+            "The first risk manager is already approved"
```
pdyraga, line 66: I'd say "The first risk manager was approved" because there is no guarantee it is still approved given that there is an unapproval process.

contracts/CoveragePool.sol
```diff
@@ -140,4 +161,26 @@ contract CoveragePool is Ownable {
...
+        external
+        view
+        returns (uint256)
+    {
+        require(
+            riskManagerApprovalTimestamps[riskManager] > 0,
+            "Risk manager approval not initiated"
+        );
+        uint256 elapsed =
+            /* solhint-disable-next-line not-rely-on-time */
```
pdyraga, line 175: I am confused. Linter is complaining but there is no public view function here. 🤔

test/CoveragePool.test.js
```diff
@@ -213,6 +293,20 @@ describe("CoveragePool", () => {
             )
           ).to.be.above(0)
         })
+
+        it("should emit RiskManagerApprovalStarted event", async () => {
```
pdyraga, line 297: s/RiskManagerApprovalStarted/RiskManagerUnapprovalStarted

pdyraga:
There is one typo in the test name mentioned above. Otherwise, it looks good to me.

Tests are failing but it is unrelated to logic changes in this PR. Let's merge https://github.com/keep-network/coverage-pools/pull/93 first, it should make unit tests in this PR green, merge in `main` and fix the typo. Then, we can 🚢 it.

test/CoveragePool.test.js
tomaszslabon: Fixed in d0c89b8f2b88677758133e85a80e456e32fdbb0b.

pdyraga (approved):

Issue #82, opened 2021-06-03
Documentation improvements


pdyraga:
Go through the contracts and make sure all functions have up-to-date and comprehensive documentation. Do any updates and reformatting needed.

PR #88, opened 2021-06-07
Check open auctions before signer bonds swap

lukasz-zimnoch:
Depends on #66 

Uniswap swap strategy should not allow performing swap in case open auctions exist. Additional collateral tokens acquired through the swap will cause the auction to offer more of the coverage pool value for the same price. This is unreasonable in terms of the economy.


contracts/SignerBondsUniswapV2.sol
```diff
@@ -75,13 +76,22 @@ contract SignerBondsUniswapV2 is ISignerBondsSwapStrategy, Ownable {
...
+    // If false, open auctions are not taken into account.
+    bool public openAuctionsCheck = true;
 
     event UniswapV2SwapExecuted(uint256[] amounts);
 
-    constructor(IUniswapV2Router _uniswapRouter, CoveragePool _coveragePool) {
+    constructor(
+        IUniswapV2Router _uniswapRouter,
+        CoveragePool _coveragePool,
+        Auctioneer _auctioneer
```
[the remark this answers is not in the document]
pdyraga: I think this cyclic dependency Tomasz mentioned is a symptom of a bigger problem with how we do swaps. Specifically, `swapSignerBonds` is not executing a swap but just depositing funds on a particular strategy. For this reason, there is no point to require no auction is open before this method is called. Secondly, what happens if most of the liquidity is removed from Uniswap v2 and governance was not able to swap all assets? ETH will be locked forever in Uniswap V2 strategy...

What I think we could do:
- Let's not call `swapSignerBonds` from the risk manager. Let's accumulate all ETH on the risk manager.
- Let's add a `withdrawBonds(uint256 amount)` function to the risk manager that can be called only by the current strategy.
- Let's not pass `_signerBondsSwapStrategy` as a constructor parameter of `CoveragePool`. We do not need the strategy to be set right away for the risk manager to work in this model, it can be set later.
- (Optional) instead of the constructor, we can `initialize` Risk Manager so that we do not have the cyclic dependency and we can approve right away.

contracts/Auctioneer.sol
```diff
@@ -30,6 +30,7 @@ contract Auctioneer is CloneFactory {
     // which will be used as a master contract for cloning.
     address public masterAuction;
     mapping(address => bool) public openAuctions;
+    uint256 public openAuctionsCount;
```
pdyraga, line 33: WDYT about adding `hasOpenAuctions` function?

contracts/SignerBondsUniswapV2.sol
```diff
@@ -75,13 +76,22 @@ contract SignerBondsUniswapV2 is ISignerBondsSwapStrategy, Ownable {
...
+    // swaps cannot be performed if open auctions count is bigger than zero.
+    // If false, open auctions are not taken into account.
+    bool public openAuctionsCheck = true;
 
     event UniswapV2SwapExecuted(uint256[] amounts);
 
-    constructor(IUniswapV2Router _uniswapRouter, CoveragePool _coveragePool) {
+    constructor(
+        IUniswapV2Router _uniswapRouter,
+        CoveragePool _coveragePool,
```
pdyraga, line 88: We can also pass `IERC20 _collateralToken, address _recipient` to reduce problems with dependencies.

contracts/SignerBondsUniswapV2.sol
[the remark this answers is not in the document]
lukasz-zimnoch: Please see the new approach: https://github.com/keep-network/coverage-pools/pull/88/commits/3f36835113dac64bdd05101fd10fd69b0c143f3a

contracts/Auctioneer.sol
lukasz-zimnoch: Done as part of https://github.com/keep-network/coverage-pools/pull/88/commits/3f36835113dac64bdd05101fd10fd69b0c143f3a

contracts/SignerBondsUniswapV2.sol
lukasz-zimnoch: I think we could leave as is due to https://github.com/keep-network/coverage-pools/pull/88/commits/3f36835113dac64bdd05101fd10fd69b0c143f3a

contracts/SignerBondsEscrow.sol
```diff
@@ -28,9 +28,19 @@ import "@openzeppelin/contracts/access/Ownable.sol";
...
-    /// @notice Swaps signer bonds.
+    /// @notice Receive ETH upon withdrawal of risk manager's signer bonds.
+    receive() external payable {}
+
+    /// @notice Swaps the given signer bonds amount from the given risk manager.
     /// @dev Adds incoming bonds to the overall contract balance.
-    function swapSignerBonds() external payable override {}
+    /// @param riskManager Address of the risk manager
+    /// @param amount Amount of signer bonds being swapped.
+    function swapSignerBonds(RiskManagerV1 riskManager, uint256 amount)
```
dimpar, line 38: Don't we need `onlyOwner` modifier here? It looks that anyone can call this now.

lukasz-zimnoch: Anyone can invoke the `swapSignerBonds` but only the owner can withdraw.

dimpar: Right, I just wonder if there's a malicious case when one can call this function and not allow to swap bonds on `SignerBondsUniswapV2` contract cause ETH are in this contract. I guess governance can call then `withdraw(SignerBondsUniswapV2.address)` but that's just an additional step for them.

lukasz-zimnoch: `SignerBondsEscrow` has nothing to do with `SignerBondsUniswapV2`. It's a separate strategy.

dimpar: Ah, I overlooked the fact that only 1 strategy can be assigned to a RiskManager at a time.

contracts/SignerBondsUniswapV2.sol
```diff
@@ -156,10 +180,17 @@ contract SignerBondsUniswapV2 is ISignerBondsSwapStrategy, Ownable {
     ///      tokens are sent to the asset pool address set during
     ///      contract construction.
     /// @param amount Amount to swap.
```
tomaszslabon, line 182: There should be one more param in the description - `riskManager`.

test/system/uniswap-v2.test.js
```diff
@@ -96,17 +118,17 @@ describeFn("System -- swap signer bonds on UniswapV2", () => {
...
-          gasLimit: 185000,
+        .swapSignerBonds(riskManagerV1.address, ethers.utils.parseEther("10"), {
+          gasLimit: 200000,
         })
     })
 
     it(
-      "should take the swapped amount from the swap strategy " +
+      "should take the swapped amount from the risk manager " +
         "contract balance",
```
tomaszslabon, line 128: This test name should be short enough to keep it in one line.

test/SignerBondsUniswapV2.test.js
```diff
@@ -150,12 +156,35 @@ describe("SignerBondsUniswapV2", () => {
...
+    context("when the caller is not the governance", () => {
+      it("should revert", async () => {
+        await expect(
+          signerBondsUniswapV2.connect(other).setOpenAuctionsCheck(false)
+        ).to.be.revertedWith("Ownable: caller is not the owner")
+      })
+    })
+
+    context("when the caller is the governance", () => {
+      it("should set the open auctions chec parameter", async () => {
```
dimpar, line 169: chec -> check

dimpar:
lgtm!

contracts/Auctioneer.sol
```diff
@@ -127,6 +134,7 @@ contract Auctioneer is CloneFactory {
         );
 
         openAuctions[cloneAddress] = true;
+        openAuctionsCount++;
```
pdyraga, line 137: Until we are not on Solidity 0.8, I think it is safer to use SafeMath `.add(1)`, `sub(1)` for increment/decrement, just in case there is a code path when the operation can overflow.

Side note, we need to standardize Solidity version in our files, I've just opened an issue for that, #98.

contracts/RiskManagerV1.sol
```diff
@@ -57,8 +57,11 @@ interface ITBTCDepositToken {
 /// @dev This interface is meant to abstract the underlying signer bonds
 ///      swap strategy and make it interchangeable for the governance.
 interface ISignerBondsSwapStrategy {
```
pdyraga, line 59: I would argue we no longer need this interface. `swapSignerBonds` is not called by other contract and every implementation could have its own signature for the swap/withdrawal function. For example, `SignerBondsEscrow`, instead of `swapSignerBonds` that is not swapping anything, would probably be good with just `withdraw`.

contracts/SignerBondsEscrow.sol
pdyraga, line 38: Do we need this function? Wouldn't it be better to call `riskManager.withdrawSignerBonds(amount);` as a part of `withdraw` and modify `withdraw` signature so that it accepts `amount` as well? And even more high level question - do we still need `SignerBondsEscrow` or maybe `SignerBondsManualSwap` next to `SignerBondsUniswapV2`? Historically, `SignerBondsEscrow` was indeed an escrow funded with each purchased signer bonds. Now, those bonds are deposited on the risk manager, until a strategy decides to withdraw them.

Right now, withdrawing bonds requires two transactions and I do not see a use case for withdrawing bonds from the risk manager but not withdrawing them from the escrow contract and accumulating them in the escrow.

contracts/SignerBondsUniswapV2.sol
```diff
@@ -75,6 +75,10 @@ contract SignerBondsUniswapV2 is ISignerBondsSwapStrategy, Ownable {
     // Determines the deadline in which the swap transaction has to be mined.
     // If that deadline is exceeded, transaction will be reverted.
     uint256 public swapDeadline = 20 minutes;
+    // Determines if the open auctions check should be performed. If true,
+    // swaps cannot be performed if open auctions count is bigger than zero.
```
pdyraga, line 79: Maybe instead of "if open auctions count is bigger than zero" we can say "if there is at least one open auction"?

contracts/SignerBondsUniswapV2.sol
```diff
@@ -75,6 +75,10 @@ contract SignerBondsUniswapV2 is ISignerBondsSwapStrategy, Ownable {
...
+    // If false, open auctions are not taken into account.
+    bool public openAuctionsCheck = true;
```
pdyraga, line 81: Maybe `revertIfAuctionOpen`? This way we are more explicit about what this flag represents, "check" is too abstract.

contracts/SignerBondsUniswapV2.sol
```diff
@@ -91,9 +95,21 @@ contract SignerBondsUniswapV2 is ISignerBondsSwapStrategy, Ownable {
...
-    function swapSignerBonds() external payable override {}
+    /// @notice Receive ETH upon withdrawal of risk manager's signer bonds.
+    /// @dev Do not send arbitrary funds. They will be locked forever.
+    receive() external payable {}
+
+    /// @notice Swaps the given signer bonds amount from the given risk manager.
+    /// @dev Swaps incoming signer bonds on Uniswap V2.
+    /// @param riskManager Address of the risk manager
+    /// @param amount Amount of signer bonds being swapped.
+    function swapSignerBonds(RiskManagerV1 riskManager, uint256 amount)
```
pdyraga, line 106: Given that this function accepts now risk manager as a parameter so it is not bound to a single implementation of a risk manager and we can use the same `SignerBondsUniswapV2` contract for, for example, both tBTC v1 and tBTC v2 risk managers, I would rather define an interface (`IWithdrawable`?) in `SignerBondsUniswapV2` with `withdrawSignerBonds` and I'd accept this interface here instead of `RiskManagerV1`.

contracts/SignerBondsUniswapV2.sol
```diff
@@ -91,9 +95,21 @@ contract SignerBondsUniswapV2 is ISignerBondsSwapStrategy, Ownable {
...
+        external
+        override
+    {
+        riskManager.withdrawSignerBonds(amount);
+        swapSignerBondsOnUniswapV2(riskManager, amount);
```
pdyraga, line 111: Instead of passing `riskManager` to `swapSignerBondsOnUniswapV2` as it has nothing to do with the swapping logic, I'd consider performing an open auction check in this function.

pdyraga (commented):
I was not reviewing tests. Leaving it for the second round.

contracts/RiskManagerV1.sol
pdyraga: On second thought, we could have this interface with e.g. `onSignerBondsPurchased` implemented empty in existing strategies and left for future, in case in a third strategy we decide to do the swap automatically, on purchase.

contracts/SignerBondsUniswapV2.sol
lukasz-zimnoch: There are several problems with the approach you propose. First, we don't know if `signer bonds` concept will be present in tBTC v2. Secondly, `IWithdrawable` is not correct as we also need to check `hasOpenAuctions` on the risk manager. We can extract an `IRiskManager` interface but having `hasOpenAuctions` here feels bad as auctions are not a risk manager concept. Also, we must have that interface in `interfaces` dir to avoid cyclic dependencies as both strategies import `ISignerBondsSwapStrategy` from `RiskManagerV1.sol`. Another option is moving the open auctions check into the risk manager itself. We could, for example, revert `withdrawSignerBonds` if there are open auctions. However, I still have a feeling we're mixing unrelated concepts here. Have you any other ideas how we can solve the problem here?

contracts/SignerBondsUniswapV2.sol
lukasz-zimnoch: https://github.com/keep-network/coverage-pools/pull/88/commits/c70beccb1628146ed51f8dd2cd4b2f9154ca43fb

test/system/uniswap-v2.test.js
lukasz-zimnoch: https://github.com/keep-network/coverage-pools/pull/88/commits/91d89730db307314c606713e113a4d42294f694a

test/SignerBondsUniswapV2.test.js
lukasz-zimnoch: https://github.com/keep-network/coverage-pools/pull/88/commits/91d89730db307314c606713e113a4d42294f694a

contracts/Auctioneer.sol
lukasz-zimnoch: https://github.com/keep-network/coverage-pools/pull/88/commits/e010a6fea5ef1f0aad862a924727c29b87fb2bf3

contracts/RiskManagerV1.sol
lukasz-zimnoch: https://github.com/keep-network/coverage-pools/pull/88/commits/fefb57fb4ce4b9d57fbd22ee4fa6e4708b2a0640

contracts/SignerBondsEscrow.sol
lukasz-zimnoch: Done in https://github.com/keep-network/coverage-pools/pull/88/commits/874455749efe522f06e634c8e36ef977a4a3cb44

contracts/SignerBondsUniswapV2.sol
lukasz-zimnoch: https://github.com/keep-network/coverage-pools/pull/88/commits/98f863ff3d68d17cb151559844177a59119aed52

contracts/SignerBondsUniswapV2.sol
lukasz-zimnoch: https://github.com/keep-network/coverage-pools/pull/88/commits/98f863ff3d68d17cb151559844177a59119aed52

contracts/SignerBondsUniswapV2.sol
lukasz-zimnoch: Resolved as part https://github.com/keep-network/coverage-pools/pull/88/commits/fefb57fb4ce4b9d57fbd22ee4fa6e4708b2a0640 . This function has been removed.

contracts/SignerBondsUniswapV2.sol
```diff
@@ -103,9 +107,13 @@ contract SignerBondsUniswapV2 is ISignerBondsSwapStrategy, Ownable {
...
-    /// @notice Swaps signer bonds.
-    /// @dev Adds incoming bonds to the overall contract balance.
-    function swapSignerBonds() external payable override {}
+    /// @notice Receive ETH upon withdrawal of risk manager's signer bonds.
+    /// @dev Do not send arbitrary funds. They will be locked forever.
+    receive() external payable {}
+
+    /// @notice Notifies the strategy about signer bonds purchase.
+    /// @param amount Amount of purchased signer bonds.
+    function onSignerBondsPurchased(uint256 amount) external override {}
```
dimpar, line 116: Is this left intentionally empty?

lukasz-zimnoch: https://github.com/keep-network/coverage-pools/pull/88#discussion_r650672931

contracts/SignerBondsUniswapV2.sol
```diff
@@ -59,12 +59,12 @@ interface IUniswapV2Pair {
 /// @title SignerBondsUniswapV2
 /// @notice ETH purchased by the risk manager from tBTC signer bonds needs to be
 ///         swapped and deposited back to the coverage pool as collateral.
-///         SignerBondsUniswapV2 is a swap strategy implementation allowing the
-///         risk manager to store purchased ETH signer bonds so that any
-///         interested part can later swap them on Uniswap v2 exchange and
-///         deposit as coverage pool collateral. The governance can set crucial
-///         swap parameters: max allowed percentage impact, slippage tolerance
-///         and swap deadline, to force reasonable swap outcomes.
+///         SignerBondsManualSwap is a swap strategy implementation which
```
pdyraga, line 62: s/SignerBondsManualSwap/SignerBondsUniswapV2

test/Auctioneer.test.js
```diff
@@ -51,7 +51,10 @@ describe("Auctioneer", () => {
   })
 
   describe("createAuction", () => {
+    let previousAuctionsCount
```
pdyraga, line 54: Instead of noting it, can't we change the main `before` into `beforeEach` so that each test is independent? I belive we use this pattern in every other test file.

contracts/test/RiskManagerV1Stub.sol
```diff
@@ -32,4 +32,8 @@ contract RiskManagerV1Stub is RiskManagerV1 {
         tbtcToken.safeTransferFrom(msg.sender, address(this), amount);
         tbtcSurplus = tbtcSurplus.add(amount);
     }
+
+    function setOpenAuctionsCount(uint256 _openAuctionsCount) external {
```
pdyraga, line 36: Is it used anywhere now?

contracts/SignerBondsUniswapV2.sol
pdyraga: >  We can extract an `IRiskManager` interface but having `hasOpenAuctions` here feels bad as auctions are not a risk manager concept.

I... think it is fine. The risk manager is opening auctions and is managing their length. It also has `onAuctionFullyFilled` and `onAuctionPartiallyFilled` so although it is not the component managing the auctions, we can say that it has open auctions - the auctions it opened to gather funds for purchasing signer bonds...

I'd vote for `IRiskManager` with `hasOpenAuctions` personally.

Another option is to accept that the swap strategy does not work with any risk manager but with `RiskManagerV1` explicitly, and make `RiskManagerV1` a constructor parameter of the strategy. This creates a chicken and egg problem though so we would need to remove swap strategy from the risk manager's constructor parameters and let to initiate the risk manager with the first swap strategy with no delay (same as we do for risk manager in coverage pool) but... this feels too complicated.

lukasz-zimnoch: Ok here it is https://github.com/keep-network/coverage-pools/pull/88/commits/5c74fa0af23afe774a52c87a3769e73fd07f2976

contracts/SignerBondsUniswapV2.sol
lukasz-zimnoch: https://github.com/keep-network/coverage-pools/pull/88/commits/caca1d5f578be58ed6ab0d2e616f4342154153c6

test/Auctioneer.test.js
lukasz-zimnoch: https://github.com/keep-network/coverage-pools/pull/88/commits/571dd99fb00e43b4e6bfb1b08beb50ecd05f7a23

contracts/test/RiskManagerV1Stub.sol
lukasz-zimnoch: Nope. Removed it https://github.com/keep-network/coverage-pools/pull/88/commits/c5b872eb5156bbf7e2dfe8d16c496505fd046f9c

lukasz-zimnoch:
@pdyraga all comments addressed!

contracts/interfaces/IRiskManager.sol
```diff
@@ -0,0 +1,26 @@
...
+//                           Trust math, not hardware.
+
+// SPDX-License-Identifier: MIT
+
+pragma solidity <0.9.0;
+
+/// @title IRiskManager
+interface IRiskManager {
+    /// @notice Withdraws the given amount of accumulated signer bonds.
+    /// @param amount Amount of signer bonds being withdrawn.
```
pdyraga, line 20: We need to improve documentation of this function, especially who can do that and in what circumstances, what is the expected behavior from the implementation. No reason to block on this one though, it can be done as a part of https://github.com/keep-network/coverage-pools/issues/82.

pdyraga (approved):
Count me in.

Issue #59, opened 2021-05-21
`receiveApproval` allows to decrease withdrawals amounts and to invest on the behalf of others


montyly:
Severity: High
Difficulty: Low

## Description
`AssetPool.receiveApproval` allows to make a deposit on the behalf of an arbitrary address, and can lead to making withdrawal worthless, or make undesired investments.

`AssetPool.receiveApproval` allows arbitrary `from`:
https://github.com/keep-network/coverage-pools/blob/4aee098edbd4755d57dcfc7649ca80ab105500bb/contracts/AssetPool.sol#L85-L97
Which will lead to execute the mint of tokens and a `transferFrom` operation:
https://github.com/keep-network/coverage-pools/blob/4aee098edbd4755d57dcfc7649ca80ab105500bb/contracts/AssetPool.sol#L257-L258

Anyone can thus invest on the behalf of users that would approve `AssetPool`.

Moreover, several high-profile tokens do not decrease the approval on `transferFrom` if the caller is the spender. If `collateralToken` is one of these tokens, `collateralToken.safeTransferFrom(depositor, address(this), amount)` will be a NOP if `depositor` is the `AssetPool`'s address. 

As a result, an attacker can mint unlimited `underwriterToken` tokens. While the minted tokens will be trapped in the `AssetPool` it will also lead the withdrawal amounts to be significantly smaller than excepted (`underwriterToken.totalSupply` will increase, but not `collateralToken.balanceOf(address(this));`):

https://github.com/keep-network/coverage-pools/blob/4aee098edbd4755d57dcfc7649ca80ab105500bb/contracts/AssetPool.sol#L167-L177

## Exploit Scenario
- Bob deploys an `AssetPool`, where the collateral token is [`UNI`](https://etherscan.io/address/0x1f9840a85d5af5bf1d1762f925bdaddc4201f984#code).
- Bob invest $100,000 worth of token. Bob has 10,000 `underwriterToken` tokens
- Later on, Bob starts a withdrawal. 
- Eve calls `AssetPool.receiveApproval`, and mints 1,000,000 `underwriterToken`.
- As a result, Bob share is worthless, and Bob receives almost nothing from his withdrawal

## Recommendations
Short term, remove `AssetPool.receiveApproval`

Long term, investigate the [token integration checklist](https://github.com/crytic/building-secure-contracts/blob/master/development-guidelines/token_integration.md), and ensure the system is robust against the ERC20 edge cases.

pdyraga:
[...]

PR #90, opened 2021-06-08
Added self-deposit safeguard

pdyraga:
Closes: #59 

Several high-profile tokens do not decrease the approval on `transferFrom` if the
caller is the spender. For example, [UNI](https://etherscan.io/address/0x1f9840a85d5af5bf1d1762f925bdaddc4201f984#code) is not even checking the allowance if
`spender == src`:
```
function transferFrom(address src, address dst, uint rawAmount) external returns (bool) {
  address spender = msg.sender;
   // (...)

  if (spender != src && spenderAllowance != uint96(-1)) {
    // update allowance
  }

  _transferTokens(src, dst, amount);
  return true;
}
```

If `collateralToken` were one of these tokens, `receiveApproval` could be used to
mint unlimited number of COV and made investments of underwriters worthless:

```
receiveApproval(
    AssetPool_address,
    UNI.balanceOf(AssetPool_address),
    UNI_address, ""
)
```
calls
```
UNI.safeTransferFrom(
    AssetPool_address,
    AssetPool_address,
    UNI.balanceOf(AssetPool_address)
)
```
in `_deposit` and `spender == src` and COV tokens can be minted.

This vulnerability was fixed in c205f56 but we are adding an additional safeguard in
`_deposit` making sure Asset Pool can not self-deposit as there is no scenario
in the system where that would happen.

This scenario could not happen in v1 of coverage pools with a single-asset pool for KEEP
but we are fixing it regardless, for future multi-asset implementation for v2.

### DIFF
--- a/contracts/AssetPool.sol
+++ b/contracts/AssetPool.sol
@@ -274,6 +274,8 @@ contract AssetPool is Ownable {
     }
 
     function _deposit(address depositor, uint256 amount) internal {
+        require(depositor != address(this), "Self-deposit not allowed");
+
         rewardsPool.withdraw();
 
         uint256 covSupply = underwriterToken.totalSupply();


contracts/AssetPool.sol
```diff
@@ -274,6 +274,8 @@ contract AssetPool is Ownable {
     }
 
     function _deposit(address depositor, uint256 amount) internal {
+        require(depositor != address(this), "Self-deposit not allowed");
```
dimpar, line 277: Do we want to add a test for this check?

pdyraga: We could and I was thinking about it as well but... calling this function so that `depositor == address(this)` is currently not possible, and it is tested [here](https://github.com/keep-network/coverage-pools/blob/c205f56836b8208c081becd31c7d08c920273ae5/test/AssetPool.test.js#L229-L242). This `require` is actually not needed in this setup but it was an audit recommendation to add it and I agree it makes sense to be very explicit about this requirement. Making this function public in a stub to test this requirement... I am not sure if it's worth it. Public interface is protected anyway.

dimpar (approved):

PR #91, opened 2021-06-08
Deploy contracts to local network

r-czajkowski:
This PR adds the required files to deploy contracts to a local network. This is required to integrate KEEP Token Dashboard with the coverage pools.

I decided to use `truffle` instead of `hardhat-deploy` plugin to deploy contracts to the local network. The main factor is that the `hardhat` library builds the contract artifact differently.

Hardhat:
```
{
 abi: [<abi here>]
 address: "<address here>"
}
```

Truffle:
```
{
 abi: [<abi here>],
 networks: {
  "1101": {
   address: "<address here>",
   transactionHash: "<tx hash here>",
  }
 }
}
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,8 @@
     "@openzeppelin/contracts": "^3.3.0"
   },
   "devDependencies": {
+    "@keep-network/keep-core": ">1.8.0-dev <1.8.0-pre",
+    "@keep-network/common.js": "0.0.1-3",
     "@nomiclabs/hardhat-ethers": "^2.0.2",
     "@nomiclabs/hardhat-waffle": "^2.0.1",
     "chai": "^4.3.4",
--- a/truffle.js
+++ b/truffle.js
@@ -1,0 +1,15 @@
+module.exports = {
+  networks: {
+    local: {
+      host: "localhost",
+      port: 8545,
+      network_id: "*",
+      websockets: true,
+    },
+  },
+  compilers: {
+    solc: {
+      version: "0.7.6", // Fetch exact version from solc-bin (default: truffle's version)
+    },
+  },
+}
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ node_modules/
 artifacts/
 cache/
 typechain/
+build/
--- a/contracts/Migrations.sol
+++ b/contracts/Migrations.sol
@@ -1,0 +1,25 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity <0.8.0;
+
+contract Migrations {
+    address public owner;
+    uint256 public last_completed_migration;
+
+    modifier restricted() {
+        if (msg.sender == owner) _;
+    }
+
+    constructor() {
+        owner = msg.sender;
+    }
+
+    function setCompleted(uint256 completed) public restricted {
+        last_completed_migration = completed;
+    }
+
+    function upgrade(address new_address) public restricted {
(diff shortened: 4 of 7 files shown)


package.json
```diff
@@ -3,6 +3,8 @@
     "@openzeppelin/contracts": "^3.3.0"
   },
   "devDependencies": {
+    "@keep-network/keep-core": ">1.8.0-dev <1.8.0-pre",
+    "@keep-network/common.js": "0.0.1-3",
```
pdyraga, line 7: Can we avoid creating this dependency? KEEP Token Dashboard should depend on coverage pools but there should be no dependency between core contracts and coverage pool contracts.

truffle.js
```diff
@@ -0,0 +1,15 @@
...
+      network_id: "*",
+      websockets: true,
+    },
+  },
+  compilers: {
+    solc: {
+      version: "0.7.6", // Fetch exact version from solc-bin (default: truffle's version)
+    },
+  },
+}
```
pdyraga, line 15: It feels wrong to have both hardhat and truffle in the same project. 😬 

What is the problem with the package structure? Can we instead implement an adapter that would support both structures in the code importing them?

cc @nkuba given that he has more experience with Hardhat deployments.

package.json
r-czajkowski: We need the KEEP token address as a `collateralToken`, right? This dependency is needed to fetch KEEP token address. We do the same in the [`keep-ecdsa`](https://github.com/keep-network/keep-ecdsa/blob/master/solidity/migrations/2_deploy_contracts.js#L39-L63) to fetch eg. `Random Beacon` address.

truffle.js
nkuba: I don't like having both hardhat and truffle as they require maintaining two different configurations and migrations scripts.

A while ago Hardhat was not producing migration artifacts containing deployment details, e.g. network, address, or transaction hash. That's why we had to create our [own tool for mainnet migrations](https://github.com/keep-network/keep-migrations). I'm not sure if anything has changed in that matter. If we don't find any plugin or hardhat's feature that supports what we need regarding produced artifacts I'm afraid we need to craft the solution on our own. The advantage of doing this is the possibility to transition from truffle to hardhat in all other modules.

nkuba: Maybe https://github.com/wighawag/hardhat-deploy is something we need?

r-czajkowski: Yes, I've seen it and played with it. The hardhat returned artifact in this format:
```
{
 abi: [<abi here>]
 address: "<address here>"
}
```

I'll dig into it and we'll see if we can add more deployment details.

mhluongo: > The advantage of doing this is the possibility to transition from truffle to hardhat in all other modules.

This would be so nice. Switching to all hardhat + ethers :yum:

pdyraga:
@lukasz-zimnoch Is it fine to assume the work done in #103 eliminates the need for this change? What about the structure of the artifact?

lukasz-zimnoch:
> @lukasz-zimnoch Is it fine to assume the work done in #103 eliminates the need for this change? What about the structure of the artifact?

Yeah I think so. Regarding structure, #103 provides all information needed by the dApp. The structure is a little bit different than Truffle artifacts but can be integrated with the dApp in a seamless way.

lukasz-zimnoch:
Closing in favor of #103

Issue #61, opened 2021-05-24
Lack of chainID validation allows re-using signatures across forks


montyly:
Severity: High
Difficulty: High

## Description
Calling `UnderwriterToken.permit` allows an asset holder to move tokens on behalf of another user by using a signature. The lack of `chainId` verification on every signature can enable an attacker to withdraw rewards multiple times if the chain is forked.

`UnderwriterToken.permit` uses `ecrecover` to allow changing the approval of users:
https://github.com/keep-network/coverage-pools/blob/4aee098edbd4755d57dcfc7649ca80ab105500bb/contracts/ERC20WithPermit.sol#L101-L146

A domain separator and the chainID are included in the signature scheme when the contract is deployed:
https://github.com/keep-network/coverage-pools/blob/4aee098edbd4755d57dcfc7649ca80ab105500bb/contracts/ERC20WithPermit.sol#L48-L61

However, this `chainID` is read at the time of deployment. In the event of a post-deployment chain hard fork, the `chainID` cannot be updated, and signatures may be replayed across both versions of the chain. As a result, an attacker could reuse signatures to receive user funds on both chains. 

## Exploit Scenario
A fork of mainnet is performed on Eth 2.0. The tokens on the old chain lose their value, but they keep the same value on the new chain. Bob allows Eve to transfer its tokens on the old chain. Eve uses the signature to also withdraw the tokens on the new chain. 

## Recommendations
Short term, add `chainID` to the signature schema to prevent post-deployment forks from affecting this permit.

Long term, identify and document the risks associated with having forks of multiple chains, and develop related mitigation strategies.

pdyraga:
I am afraid I am not entirely following and I need some more explanation. Is this issue only about Ethereum 1.0 -> Ethereum 2.0 upgrade or about Ethereum 1.0 chain forks as well?

Once ready, Ethereum 1.0 will merge with Ethereum 2.0 Beacon Chain, becoming its own shard so the state of the contract will be maintained.  Eve can not replay permit transactions because Ethereum 2.0 will expect a higher nonce in the Permit structure.

Thinking about two different chains with two different IDs, Eve can not replay permit transactions because of a different chain ID in DOMAIN_SEPARATOR.

In the case of forks at the same chain, say Ethereum 1.0 chain forks, transactions can be replayed between forks but that is always true, not only in case of `ERC20WithPermit`.

montyly:
[...]

PR #92, opened 2021-06-08
Validate chain ID for every permit call

pdyraga:
Closes: #61

Calling `UnderwriterToken.permit` allows an asset holder to move tokens on behalf
of another user by using a signature. The lack of chain ID verification on every
signature could enable an attacker to withdraw rewards multiple times on behalf
of someone else if the chain was forked with a different chain ID.

The chain ID was read only at the deployment time and `DOMAIN_SEPARATOR` was
constructed based on that value. It could happen that the forked chain had
another chain ID, and given that we were not performing chain ID verification
for each permit call, some transactions could be replayed. The same issue
is in LP token v2 version this code is based on.

To address this problem, chain ID and `DOMAIN_SEPARATOR` constructed
for the given chain ID are cached and chain ID is validated for every permit
call. In case the cached value does not match the current one, a new
`DOMAIN_SEPARATOR` is constructed based on the current chain ID value.

### DIFF
--- a/contracts/ERC20WithPermit.sol
+++ b/contracts/ERC20WithPermit.sol
@@ -33,23 +33,24 @@ import "./interfaces/IERC20WithPermit.sol";
 contract ERC20WithPermit is Ownable, IERC20WithPermit {
     using SafeMath for uint256;
 
-    uint8 public constant decimals = 18;
-
-    string public name;
-    string public symbol;
-
-    uint256 public override totalSupply;
     mapping(address => uint256) public override balanceOf;
     mapping(address => mapping(address => uint256)) public override allowance;
 
-    uint256 public cachedChainId;
-    bytes32 public cachedDomainSeparator;
+    mapping(address => uint256) public override nonces;
+
+    uint256 public immutable cachedChainId;
+    bytes32 public immutable cachedDomainSeparator;
 
     // keccak256("Permit(address owner,address spender,uint256 value,uint256 nonce,uint256 deadline)");
     bytes32 public constant override PERMIT_TYPEHASH =
         0x6e71edae12b1b97f4d1f60370fef10105fa2faae0126114a169c64845d6126c9;
 
-    mapping(address => uint256) public override nonces;
+    uint256 public override totalSupply;
+
+    string public name;
+    string public symbol;
+
+    uint8 public constant decimals = 18;
 
     constructor(string memory _name, string memory _symbol) {
         name = _name;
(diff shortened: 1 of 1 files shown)


contracts/ERC20WithPermit.sol
```diff
@@ -42,32 +42,21 @@ contract ERC20WithPermit is Ownable, IERC20WithPermit {
     mapping(address => uint256) public override balanceOf;
     mapping(address => mapping(address => uint256)) public override allowance;
 
-    bytes32 public override DOMAIN_SEPARATOR;
+    uint256 public cachedChainId;
```
dimpar, line 45: minor, but can we move this below `totalSupply` for gas efficiency?

contracts/ERC20WithPermit.sol
```diff
@@ -177,6 +166,15 @@ contract ERC20WithPermit is Ownable, IERC20WithPermit {
         _burn(account, amount);
     }
 
+    /* solhint-disable-next-line func-name-mixedcase */
+    function DOMAIN_SEPARATOR() public view override returns (bytes32) {
+        if (chainId() == cachedChainId) {
+            return cachedDomainSeparator;
+        } else {
+            return buildDomainSeparator();
```
dimpar, line 174: in case `chainId() =/= cachedChainId` we build a new domain separator. But do we also want to update cachedChainId, so the next time we don't need to build it again assuming chain id doesn't change that frequently..

pdyraga: We don't because in practice, this situation should never happen and `cachedChainId` should be marked as `immutable` for gas efficiency which I forgot to do. So glad you brought it up! 

Please see 012346fdafebca91cc4504912c637d4f0daf701e

contracts/ERC20WithPermit.sol
pdyraga: Yes, let's group them together. Please see 012346fdafebca91cc4504912c637d4f0daf701e.

contracts/ERC20WithPermit.sol
dimpar: okay, although I still remain a bit confused. Despite that this situation should never happen, yet we still have this check and updating cache within this validation seems reasonable. Anyway, I take that there is a veeery tiny chance it might happen and that's why we have it here. I'm ok leaving as-is.

dimpar (approved):

contracts/ERC20WithPermit.sol
pdyraga: If we set the field as `immutable`, it will be cheaper to access it but also we cannot modify it later. Given that in practice we should never need to update it, we sacrifice gas efficiency updates but for the unlikely case.

Issue #83, opened 2021-06-03
System tests


pdyraga:
We have a set of system tests working on Hardhat Network on a forked mainnet chain. 
At the start, the fork is being reset to the specific starting block which determines the initial test state. Tests use real TBTC token contract and Deposit contract. Currently, the following scenarios are covered:

- Liquidation happy path with a fully-filled auction. Deposit is in a liquidation state, bidder takes the entire auction. We assert that the deposit has been liquidated, auction is closed, signer bonds are processed, and reasonable amount of gas is consumed.
- Deposit validation. It is not possible to open an auction for a fake deposit.
- Uniswap V2 ETH->KEEP bonds swap happy path. We assert signer bonds are properly swapped and sent to the asset pool. We assert a reasonable amount of gas is consumed.

This issue aims at identifying additional system test scenarios and implementing them.

pdyraga:
Proposed new scenarios:
- Auction fully filled in one shot. Assert the correct amount has been seized from the pool.
- Auction fully filled but using multiple partial take offers. Assert the correct amount has been seized from the pool.
- Auction liquidated outside of cov pools. Assert it is not possible to take an auction (fully filled and partial offer).

... define more please.

tomaszslabon:
Perhaps this scenario:
- Auction fully filled after `auctionLength`. Ensure the whole pool was seized and transferred to the bidder.

dimpar:
All of the above scenarios were added in the table below except the last one, which I think overlaps with other tests. A test scenario was added covering buying deposit with surplus funds.

For more tests scenarios I propose to fill out this table to track what we already have and what should be added.

| Testing scenario                                            | Assertion                                                                                                                                                                                                    | Is implemented |
|-------------------------------------------------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|----------------|
| Liquidation happy path with a fully-filled auction          | Deposit is in a liquidation state, bidder takes the entire auction. We assert that the deposit has been liquidated, auction is closed, signer bonds are processed, and reasonable amount of gas is consumed. | Yes            |
| Deposit validation                                          | It is not possible to open an auction for a fake deposit.                                                                                                                                                    | Yes            |
| Uniswap V2 ETH->KEEP bonds swap happy path                  | Signer bonds are properly swapped and sent to the asset pool. We assert a reasonable amount of gas is consumed.                                                                                              | Yes            |
[...]

PR #94, opened 2021-06-10
[Integration tests] Taking offer on auctions

tomaszslabon:
[Probably not needed - do not merge]

This pull request adds three integration tests:
- `when auction fully filled before auction length elapsed`. Scenario: Auction fully filled in one shot before the auction length elapsed. Assert the correct amount has been seized from the pool and what the bidder received. 
-  `when auction fully filled after auction length elapsed`.Scenario: Auction fully filled after the auction length elapsed. Assert the whole pool was seized and transferred to the bidder.
- `when auction fully filled with multiple take offers`.Scenario: Auction fully filled but using multiple partial take offers. Assert the correct amount has been seized from the pool and what bidders received.

### DIFF
--- a/test/integration/liquidation.test.js
+++ b/test/integration/liquidation.test.js
@@ -1,33 +1,49 @@
 const { expect } = require("chai")
-const { to1e18 } = require("../helpers/contract-test-helpers")
+const {
+  to1e18,
+  increaseTime,
+  to1ePrecision,
+} = require("../helpers/contract-test-helpers")
 const { deployMockContract } = require("@ethereum-waffle/mock-contract")
 const Auction = require("../../artifacts/contracts/Auction.sol/Auction.json")
 const ITBTCDepositToken = require("../../artifacts/contracts/RiskManagerV1.sol/ITBTCDepositToken.json")
+const { BigNumber } = ethers
 
 describe("Integration -- liquidation", () => {
   const auctionLength = 86400 // 24h
   const lotSize = to1e18(10)
   const bondedAmount = to1e18(150)
   const bondAuctionThreshold = 100
+  const precision = to1ePrecision(1, 16) // 0.01
 
   let tbtcToken
+  let collateralToken
+  let underwriterToken
   let signerBondsSwapStrategy
+  let assetPool
   let coveragePool
   let riskManagerV1
   let tbtcDeposit
 
   let owner
   let bidder
+  let otherBidder
   let thirdParty
+  let rewardsManager
 
   beforeEach(async () => {
     owner = await ethers.getSigner(0)
     bidder = await ethers.getSigner(1)
-    thirdParty = await ethers.getSigner(2)
+    otherBidder = await ethers.getSigner(2)
+    thirdParty = await ethers.getSigner(3)
+    rewardsManager = await ethers.getSigner(4)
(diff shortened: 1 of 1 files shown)


test/integration/liquidation.test.js
```diff
@@ -105,6 +138,188 @@ describe("Integration -- liquidation", () => {
...
+      // velocityPoolDepletingRate = 1.0 (initial value)
+      // amountOnOffer = (timeSinceAuctionStart - startTimeOffset)
+      //   * velocityPoolDepletingRate / auctionLength =
+      //   (4320 - 0) * 1.0 / 86400 = 0.05
+      // portionToSeize = amountOnOffer * amountToTransfer / amountOutstanding =
+      //   0.05 * (4 * 10^18) / (10 * 10^18 ) = 0.02
+      // amountInAssetPool = 100 * 10^18
+      // amountToSeize = amountInAssetPool * portionToSeize =
+      //   (100 * 10^18) * 0.02 = 2 * 10^18
+      // Bidder should receive (2 * 10^18) of collateral tokens
```
tomaszslabon, line 215: Posting all these calculations here feels like an overkill, but I wanted to show how we arrived with the values of collateral tokens seized from asset pool that we expect to get in test cases.

pdyraga: > Alternatively, we could just check that after the auction was fully filled, the total number of collateral tokens the bidders received is equal to the initial amount of collateral tokens in the asset pool.

Plus, we could check that what they received is what the auction offered, especially if bidding through `AuctionBidder.takeOfferWithMin`. 

I do agree those calculations are overkill, I would leave them for unit tests.

lukasz-zimnoch (approved):
LGTM! @tomaszslabon please address the remaining comment and I think we can merge.

lukasz-zimnoch:
Please note the changes introduced by https://github.com/keep-network/coverage-pools/pull/88 before merging. It changed the way signer bonds are handled.

tomaszslabon:
@pdyraga 
As we modified the [system test scenarios](https://github.com/keep-network/coverage-pools/issues/83), I think this PR is no longer needed and should be closed without merging.

Testcase `when auction fully filled before auction length elapsed` will be covered by scenario 1. from the table (the system test already exists but will need to be updated).

Testcase `when auction fully filled with multiple take offers` will be covered by scenario 4 (we decided to simplify the flow - take offers after auction elapsed with two bidders, the scenario was too complicated for a system test). The person who takes scenario 4. will decide whether to write an integration test or a system test.

Testcase `when auction fully filled after auction length elapsed` will also be covered by scenario 4.

What do you think?

pdyraga:
This PR is going to be replaced by #114, correct @tomaszslabon ?

tomaszslabon:
> This PR is going to be replaced by #114, correct @tomaszslabon ?

@pdyraga 
Yes, it will be replaced by #114 (I'm finishing it in a moment - encountered some problem with calculations), and by another PR that adds a system test number 4 from [this list](https://github.com/keep-network/coverage-pools/issues/83#issuecomment-860604097).

lukasz-zimnoch:
Closing in favor of https://github.com/keep-network/coverage-pools/pull/114

Issue #95, opened 2021-06-11
Multiple auctions per deposit


mhluongo:
Another one! Let's reproduce this in a test before patching :pray: 

> it appears to be possible to notify a deposit being liquidated multiple times which results in the creation of multiple auctions for a single deposit.

@samczsun

PR #96, opened 2021-06-11
[System tests] Adding tests when notifying on liquidated auction

dimpar:
Creating a system test that introduces 2 bidders. The first one that takes a partial offer on an opened cov pool auction. And the second bidder which buys a deposit outside the cov pool. Testing functionalities of `notifyLiquidated()` after the deposit is bought outside the cov pool by a third party bidder.

Refs https://github.com/keep-network/coverage-pools/issues/83

Depends on: https://github.com/keep-network/coverage-pools/pull/113

PR #97, opened 2021-06-11
Risk manager does not allow to be notified twice about the same deposit

pdyraga:
Closes: #95 

It was possible to notify risk manager about the liquidation of the same
deposit twice and opening multiple auctions. Fixing it now.

### DIFF
--- a/contracts/RiskManagerV1.sol
+++ b/contracts/RiskManagerV1.sol
@@ -162,15 +162,15 @@ contract RiskManagerV1 is Auctioneer, Ownable {
             "Address is not a deposit contract"
         );
 
-        require(
-            depositToAuction[depositAddress] == address(0),
-            "Already notified on the deposit in liquidation"
-        );
-
         IDeposit deposit = IDeposit(depositAddress);
         require(
             deposit.currentState() == DEPOSIT_LIQUIDATION_IN_PROGRESS_STATE,
             "Deposit is not in liquidation state"
+        );
+
+        require(
+            depositToAuction[depositAddress] == address(0),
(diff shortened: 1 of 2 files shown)


contracts/RiskManagerV1.sol
```diff
@@ -162,6 +162,11 @@ contract RiskManagerV1 is Auctioneer, Ownable {
             "Address is not a deposit contract"
         );
 
+        require(
+            depositToAuction[depositAddress] == address(0),
+            "Already notified about the deposit"
```
dimpar, line 167: Not biggy and up to you:
- how about: "Already notified on the deposit in liquidation"
- if the flow goes through bying with surplus, then technically we still can go through this check > 1 but it will be stopped in `currentState()` check. So my proposition is to put this check below the `curretnState()` check.

pdyraga: Good points, updated!

dimpar:
lgtm but tests exceed timeout. I've increased it, but it's in this [PR](https://github.com/keep-network/coverage-pools/pull/96/files)

lukasz-zimnoch:
> lgtm but tests exceed timeout. I've increased it, but it's in this [PR](https://github.com/keep-network/coverage-pools/pull/96/files)

That happens sometimes when there are problems with ETH node connectivity. Increasing the timeout may not help in all cases. I think we must live with it :shrug:

pdyraga:
I've rerun system tests for now, it's green here @lukasz-zimnoch @dimpar

dimpar (approved):

PR #103, opened 2021-06-15
Contracts deployment script

lukasz-zimnoch:
This pull request introduces a deployment script based on the [hardhat-deploy](https://github.com/wighawag/hardhat-deploy) plugin.

To deploy all contracts on the given network, the following command can be used:
```
yarn deploy --network <network>
```

If contracts haven't been built yet or changes occurred, this task will build the contracts before running the deployment script. Once the deployment terminates, a new `deployments` directory containing all deployment info will be created. It can be directly used by dApps or other client code as it contains deployment details like chain ID, transaction hash, ABI or address for each contract. There is also a possibility to run the deployment with export mode enabled:
```
yarn deploy --network <network> --export ./export.json
```
This command produces a handy summary file containing deployment info for all contracts in one place. However, it doesn't contain the deployment transaction hash making it inappropriate for some use cases relying on this field.

The deployment script deals with external contract dependencies using environment variables. In case an address is not provided, the script deploys a stub contract. This makes the deployment easier as the script doesn't depend on any external packages and allows to obtain external addresses in the most suitable way. On the other hand, using stubs as defaults make the local development easier as there is no need to orchestrate all external dependencies. Specific env variables are:

- `KEEP_TOKEN_ADDRESS`: Determines the address of KEEP token contract
- `TBTC_TOKEN_ADDRESS`: Determines the address of TBTC token contract
- `TBTC_DEPOSIT_TOKEN_ADDRESS`: Determines the address of TDT token contract
- `UNISWAP_ROUTER_V2_ADDRESS`: Determines the address of Uniswap v2 router
- `INITIAL_SWAP_STRATEGY`: Allows setting the initial swap strategy which will be used by the risk manager. This should         be the name of one of the `ISignerBondsSwapStrategy` implementations.

The deployment script is divided into multiple sub-scripts placed in the `deploy` directory. It uses the [tags and dependencies](https://github.com/wighawag/hardhat-deploy#deploy-scripts-tags-and-dependencies) feature provided by the `hardhat-deploy` plugin. Such a structure allows running arbitrary parts of the entire deployment by using the tag mechanism. For example, to deploy only the `AssetPool` contract (with their dependencies), the following command can be used:
```
yarn deploy --network localhost --tags AssetPool
```
Multiple deployment sub-scripts also improves the readability and allows specifying dependencies between components in an explicit way.


r-czajkowski (approved):

README.adoc
```diff
@@ -95,4 +95,54 @@ There are multiple test scenarios living in the `test` directory.
...
+yarn deploy --network <network>
+```
+
+If contracts haven't been built yet or changes occurred, this task will build
+the contracts before running the deployment script. Once the deployment
+terminates, a new `deployments` directory containing all deployment info
+will be created. It can be directly used by dApps or other client code as it
+contains deployment details like chain ID, transaction hash, ABI or address for
+each contract. There is also a possibility to run the deployment with export
+mode enabled:
```
nkuba, line 113:
```suggestion
each contract. 

==== Export mode

There is also a possibility to run the deployment with export mode enabled:
```

README.adoc
```diff
@@ -95,4 +95,54 @@ There are multiple test scenarios living in the `test` directory.
...
+```
+yarn deploy --network <network> --export ./export.json
+```
+which produces a handy summary file containing deployment info for all
+contracts in one place. However, it doesn't contain the deployment transaction
+hash making it inappropriate for some use cases relying on this field.
+
```
nkuba, line 120:
```suggestion

==== External dependencies

```
```

File: README.adoc [hunk 3]
```diff
@@ -95,4 +95,54 @@ There are multiple test scenarios living in the `test` directory.
...
+external dependencies. Specific env variables are:
+
+- `KEEP_TOKEN_ADDRESS`: Determines the address of KEEP token contract
+- `TBTC_TOKEN_ADDRESS`: Determines the address of TBTC token contract
+- `TBTC_DEPOSIT_TOKEN_ADDRESS`: Determines the address of TDT token contract
+- `UNISWAP_ROUTER_V2_ADDRESS`: Determines the address of Uniswap v2 router
+- `INITIAL_SWAP_STRATEGY`: Allows setting the initial swap strategy which will
+  be used by the risk manager. This should be the name of one of the
+  `ISignerBondsSwapStrategy` implementations.
+
```
nkuba, line 136:
```suggestion

=== Deployment scripts structure and tags

```

deploy/00_resolve_keep_token.js
```diff
@@ -0,0 +1,22 @@
...
+  const keepTokenAddress = process.env.KEEP_TOKEN_ADDRESS
+  const { deploy, save, log } = deployments
+  const { deployer } = await getNamedAccounts()
+
+  if (keepTokenAddress) {
+    log(`using externally provided KeepToken address ${keepTokenAddress}`)
+
+    // Save as simple deployment just to make it accessible for next scripts.
+    await save("KeepToken", { address: keepTokenAddress })
+  } else {
```
nkuba, line 11: What do you think about adding a network check? We shouldn't deploy stubs for networks other than local (e.g. ropsten, mainnet). We should fail for non-local if the external address is missing.

nkuba (changes requested):

deploy/00_resolve_keep_token.js
r-czajkowski: Makes sense. What about the `UniswapV2Router` eg. for the `ropsten` or any other test network? Should we deploy a stub or use a real instance?

pdyraga: Let's use real instances on public testnets, if possible.

deploy/06_deploy_rewards_pool.js
```diff
@@ -0,0 +1,15 @@
+module.exports = async ({ getNamedAccounts, deployments }) => {
+  const { deploy } = deployments
+  const { deployer, rewardManager } = await getNamedAccounts()
+  const KeepToken = await deployments.get("KeepToken")
+  const AssetPool = await deployments.get("AssetPool")
+
+  await deploy("RewardsPool", {
```
nkuba, line 7: Do we need a separate script to deploy `RewardsPool`? An instance of `RewardsPoll` is already deployed as part of the `AssetPool` constructor
https://github.com/keep-network/coverage-pools/blob/34c3d2f6bdb1c608a755dae07f39d29c52297126/contracts/AssetPool.sol#L125

r-czajkowski: Let's say it's a workaround to get the `RewardsPool` contract abi in the `deployments` dir. The hardhat splits the compiled and deployed contracts artifacts between different dirs. In this repo hardhat saves compiled artifacts to `./build` but deployed artifacts to `./deployments/<network>` and we create a symbolic link only to the `deployments/<network>` dir.

nkuba: So the `RewardsPool` contract will be deployed twice, first in the `AssetPool` constructor and second in the `06_deploy_rewards_pool.js`. But we will use the address from `AssetPool` instead of the deployed artifact. This is very likely to cause a confusion.

What if we remove the `06_deploy_rewards_pool.js` script and expect the contract to be deployed inside the `AssetPool` constructor. We will obtain the address easily with a call to `AssetPool.rewardsPool`. For the `RewardsPool` ABI we will have to use compilation artifacts from the `./build` directory.

r-czajkowski: Please see b1cdc5e.

deploy/06_deploy_coverage_pool.js
```diff
@@ -0,0 +1,24 @@
...
+  })
+
+  const assetPool = await ethers.getContractAt("AssetPool", AssetPool.address)
+  const underwriterToken = await ethers.getContractAt(
+    "UnderwriterToken",
+    UnderwriterToken.address
+  )
+
+  await assetPool.transferOwnership(CoveragePool.address)
+  await underwriterToken.transferOwnership(AssetPool.address)
```
nkuba, line 20: The transactions will fail if we run the deployments scripts again e.g. to retry after some failure. We need to check the owner first.

```suggestion
  if ((await assetPool.owner()) !== CoveragePool.address) {
    await assetPool.transferOwnership(CoveragePool.address)
  }

  if ((await underwriterToken.owner()) !== AssetPool.address) {
    await underwriterToken.transferOwnership(AssetPool.address)
  }
```

nkuba: To be consistent with how we handle transaction execution in `11_approve_first_risk_manager.js` instead of `ethers` we could use what `deployments` have:
```js
  const { read, execute } = deployments
// ...
  if (
    (await read("AssetPool", { from: deployer }, "owner")) !==
    CoveragePool.address
  ) {
    await execute(
      "AssetPool",
      { from: deployer },
      "transferOwnership",
      CoveragePool.address
    )
  }

  if (
    (await read("UnderwriterToken", { from: deployer }, "owner")) !==
    AssetPool.address
  ) {
    await execute(
      "UnderwriterToken",
      { from: deployer },
      "transferOwnership",
      AssetPool.address
    )
  }
```

deploy/11_approve_first_risk_manager.js
```diff
@@ -0,0 +1,17 @@
...
+  const RiskManagerV1 = await deployments.get("RiskManagerV1")
+
+  log(`approving first risk manager ${RiskManagerV1.address}`)
+
+  await execute(
+    "CoveragePool",
+    { from: deployer },
+    "approveFirstRiskManager",
+    RiskManagerV1.address
+  )
```
nkuba, line 13: We need to check if the risk manager is already approved so we won't get a revert on deployment retry.

```suggestion
  const isRiskManagerApproved = await read(
    "CoveragePool",
    { from: deployer },
    "approvedRiskManagers",
    RiskManagerV1.address
  )

  if (!isRiskManagerApproved) {
    log(`approving first risk manager ${RiskManagerV1.address}`)

    await execute(
      "CoveragePool",
      { from: deployer },
      "approveFirstRiskManager",
      RiskManagerV1.address
    )
  } else {
    log(`risk manager ${RiskManagerV1.address} is already approved`)
  }
```

deploy/05_deploy_asset_pool.js
```diff
@@ -0,0 +1,40 @@
...
+      receipt,
+      transactionHash: AssetPool.transactionHash,
+    },
+    RewardsPool
+  )
+
+  log("The RewardsPool has been deployed in the same transacion as AssetPool.")
+  log(
+    `The RewardsPool address is: ${rewardsPoolAddress} - saving the RewardsPool deployments info.`
+  )
```
nkuba, line 34:
```suggestion
  log(`RewardsPool has been deployed at ${rewardsPoolAddress} in the same transacion as AssetPool`)
```

deploy/05_deploy_asset_pool.js
```diff
@@ -0,0 +1,40 @@
...
+    log: true,
+  })
+
+  const assetPool = await ethers.getContractAt("AssetPool", AssetPool.address)
+  const rewardsPoolAddress = await assetPool.rewardsPool()
+
+  // The`RewardsPool` contract is created in the `AssetPool` constructor so
+  // it's created in the same transaction.
+  const receipt = AssetPool.receipt
+  receipt.contractAddress = rewardsPoolAddress
```
nkuba, line 20: Is there a reason we modify the receipt? IMHO we should leave `contractAddress` unchanged as this was the original transaction.

nkuba (changes requested):

deploy/00_resolve_keep_token.js
```diff
@@ -0,0 +1,24 @@
...
+  const keepTokenAddress = process.env.KEEP_TOKEN_ADDRESS
+  const { deploy, save, log } = deployments
+  const { deployer } = await getNamedAccounts()
+
+  if (keepTokenAddress) {
+    log(`using externally provided KeepToken address ${keepTokenAddress}`)
+
+    // Save as simple deployment just to make it accessible for next scripts.
+    await save("KeepToken", { address: keepTokenAddress })
+  } else if (hre.network.name !== "local") {
```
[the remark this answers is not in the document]
nkuba: Wow, there is an even better feature we could use: https://github.com/wighawag/hardhat-deploy/tree/master#tags!
So in hardhat config we have to add `tags: ["local"],` for `hardhat` and `local` networks:
```js
  networks: {
    hardhat: {
      forking: {
// ...
      },
      tags: ["local"],
    },
    local: {
      url: "http://localhost:8545",
      tags: ["local"],
    },
  },
```
And here we do:

```suggestion
  } else if (!hre.network.tags.local) {
```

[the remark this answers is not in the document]
nkuba: We will handle this in https://github.com/keep-network/coverage-pools/pull/151.

deploy/03_resolve_uniswap_v2_router.js
```diff
@@ -0,0 +1,24 @@
...
+  const { deployer } = await getNamedAccounts()
+
+  if (uniswapV2RouterAddress) {
+    log(
+      `using externally provided UniswapV2Router address ${uniswapV2RouterAddress}`
+    )
+
+    // Save as simple deployment just to make it accessible for next scripts.
+    await save("UniswapV2Router", { address: uniswapV2RouterAddress })
+  } else {
```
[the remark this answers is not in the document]
nkuba: We will handle this in https://github.com/keep-network/coverage-pools/pull/151.

deploy/05_deploy_asset_pool.js
nkuba: I updated it in e1020703a1f8dbfacd499c1d88d9ef546623a680. Please let me know if this shouldn't be changed.

deploy/05_deploy_asset_pool.js
nkuba: 89dbd8c8616574b10956feecbe3802bc203e5908

deploy/11_approve_first_risk_manager.js
nkuba: 503a296c6685a4796638fc7a988779dc6cd4ef28

deploy/06_deploy_coverage_pool.js
nkuba: 503a296c6685a4796638fc7a988779dc6cd4ef28

nkuba (approved):
@r-czajkowski I let myself to commit updates to this PR as I had them already made locally. These are also required by other PRs we have comming in the line.

deploy/06_deploy_coverage_pool.js
```diff
@@ -1,23 +1,35 @@
...
     args: [AssetPool.address],
     log: true,
   })
 
-  const assetPool = await ethers.getContractAt("AssetPool", AssetPool.address)
-  const underwriterToken = await ethers.getContractAt(
-    "UnderwriterToken",
-    UnderwriterToken.address
-  )
+  if ((await read("AssetPool", "owner")) !== CoveragePool.address) {
```
r-czajkowski, line 12: Shouldn't we check addresses in a more secure way? I mean we should compare checksum addresses, unless we are sure that addresses are in the same "format". https://docs.ethers.io/v5/api/utils/address/#utils-getAddress.

```
ethers.utils.getAddress((await read("AssetPool", "owner")) !== ethers.utils.getAddress(CoveragePool.address)
```

r-czajkowski (approved):
LGTM! Just left a tiny note.

deploy/06_deploy_coverage_pool.js
nkuba: Great point! I will handle it in a follow-up PR.

scripts/install.sh
```diff
@@ -0,0 +1,82 @@
...
+
+LOG_START='\n\e[1;36m'  # new line + bold + cyan
+LOG_END='\n\e[0m'       # new line + reset
+DONE_START='\n\e[1;32m' # new line + bold + green
+DONE_END='\n\n\e[0m'    # new line + reset
+
+COVERAGE_POOL_PATH=$(realpath $(dirname $0)/../)
+
+# Defaults, can be overwritten by env variables/input parameters
+NETWORK_DEFAULT="local"
```
r-czajkowski, line 12: Because of f12c0ec we need to change the default network as well. @nkuba

deploy/06_deploy_coverage_pool.js
nkuba: https://github.com/keep-network/coverage-pools/pull/153

scripts/install.sh
nkuba: https://github.com/keep-network/coverage-pools/pull/153

PR #104, opened 2021-06-16
[System tests] Extracting contract deployment for system tests

dimpar:
Currently we have duplicate code which deploys the same contracts for the system tests. To make it cleaner, this change extracts common code for contract deployments and introduces contract constants for tests.
Existing tests are refactored according to the above changes. Future system tests should reuse proposed functionality.

These changes are refactoring only, no new system tests were added.


lukasz-zimnoch (commented):
Looks good but let's hold off with merging until all system tests PRs land on main.

dimpar:
If we wait on all other system tests PRs to land on main, then we will have even more files in conflict to refactor in the future. Some of the comments in the pending PRs can be resolved automatically with this guy. Plus some new tests might be added parallel to this PR. I'd rather actually merge this guy first to setup the structure for all other system tests to rely on.

pdyraga:
I'd also love to merge this one before other system tests land. We will have some conflicts but the review process should be much easier.

dimpar:
> I'd also love to merge this one before other system tests land. We will have some conflicts but the review process should be much easier.

lukasz-zimnoch (approved):

pdyraga:
What about `test/system/deposit-liquidated-externally.test.js`?

dimpar:
> What about `test/system/deposit-liquidated-externally.test.js`?

Slipped from the radar.. pls see https://github.com/keep-network/coverage-pools/pull/113

Issue #89, opened 2021-06-07
Simplify withdrawal mechanism?


pdyraga:
As currently implemented, the coverage pool withdrawal mechanism works as follows:
- Underwriter deposited into the pool. Their KEEP is in the pool earning rewards and is a subject of a potential claim.
- Underwriter initiated withdrawal. Their KEEP is in the pool earning rewards and is a subject of a potential claim.
- Underwriter is waiting for 2 weeks withdrawal delay.  Their KEEP is in the pool earning rewards and is a subject of a potential claim.
- The 2-week withdrawal delay passed. There is 1 week graceful withdrawal time for which the underwriter can complete the withdrawal. For the entire withdrawal delay time, until the underwriter completes the withdrawal, their KEEP is in the pool earning rewards and is a subject of a potential claim.
- In the case underwriter forgot (or decided not to) complete the withdrawal during the graceful withdrawal time, they have 9 weeks (63 days) before the hard withdrawal timeout. For all that time, until the underwriter completes the withdrawal, their KEEP is in the pool earning rewards and is a subject of a potential claim. For all that time. until the underwriter completes the withdrawal, KEEP and rewards accrued are getting slowly seized by the pool, proportionally to the time passed since the graceful withdrawal time ended.
- If they do not complete withdrawal after 63 days, they lose all of their tokens - they are seized by the pool.

AAVE approach for unstaking is:
- Activate cooldown, wait for the cooldown period, currently 10 days
- Complete unstaking after the cooldown period. One has 2 days for that (governable). If they do not complete within 2 days, they need to activate cooldown again and wait for the cooldown period again.

Both approaches have their pros and cons. 

What I personally do not like about the approach with "returning to normal staking" compared to the current implementation, is that in case the withdrawal has not been completed after the cooldown period, for the entire two-day unstake period, underwriters free-ride the system. They earn rewards but can withdraw at any time, especially before a claim. Smart underwriters could be activating the unstake in a loop. They cannot withdraw for 10 days but can withdraw at any time for 2 days, so they got 16% of the staking time as risk-free. On the other hand, the same situation is for our current implementation except that they need to execute two more transactions that do cost gas: withdraw and deposit again. It is not cost-free but there might be a situation when the low gas cost makes this attack more profitable.

[...]

PR #105, opened 2021-06-16
Governance improvements

pdyraga:
Depends on: #100 
Refs #89

Along with the change simplifying the withdrawal flow (see #100), we decided to let the governance alter withdrawal parameters (delay and timeout) but in such a way that if underwriters do not agree with the change, they have a chance to withdraw their positions before the change takes effect. The total delay for the change to be applied is the entire withdrawal delay period, withdrawal timeout period, and an additional 2 days to get notified about the change and make a decision about possibly withdrawing the position. This PR makes withdrawal delay and withdrawal timeout governable with the mentioned restriction.

Additionally, it changes the risk manager approval governance delay to use the same value as the asset pool. This delay is enough for all underwriters to withdraw their positions in case they do not agree with the change. Using a hardcoded, 30 days delay was fine as long as the timeout/delay for withdrawal were hardcoded in the asset pool. Now, given that they can be altered, the coverage pool can not use a hardcoded value.

Last but not least, we now allow the governance to disapprove the risk manager immediately. From the coverage pool perspective, all changes that could negatively affect underwriter's positions should be protected by a governance delay. Unapproving risk manger can only have a positive effect on underwriters positions (no longer a risk of being liquidated from the given manager), and this change can be applied immediately. If there are additional requirements needed so that e.g. peg is not broken, risk manager unapproval can be proxied through a worklock.

TODO:
- [x] improve Solidity function documentation
- [x] update `docs/design.adoc`


contracts/AssetPool.sol
```diff
@@ -296,6 +324,80 @@ contract AssetPool is Ownable, IAssetPool {
...
+    {
+        return
+            getRemainingChangeTime(
+                withdrawalTimeoutChangeInitiated,
+                withdrawalGovernanceDelay()
+            );
+    }
+
+    /// @notice The time it takes to initiate and complete the withdrawal from
+    ///         the pool plus 2 days. This governance delay should be used for
```
dimpar, line 395: Maybe it's worth adding a note why exactly 2 days?

contracts/CoveragePool.sol
```diff
@@ -185,60 +146,63 @@ contract CoveragePool is Ownable {
...
     /// @param delay Governance delay.
-    /// @param errorMsg Revert message when change not initiated
     /// @return Remaining time in seconds.
-    function getRemainingChangeTime(
-        uint256 changeTimestamp,
-        uint256 delay,
-        string memory errorMsg
-    ) internal view returns (uint256) {
-        require(changeTimestamp > 0, errorMsg);
+    function getRemainingChangeTime(uint256 changeTimestamp, uint256 delay)
```
dimpar, line 200: This function looks almost identical to the one in asset pool. Thoughts on extracting it to a lib maybe or call `assetPool.getRemainingChangeTime()`?

contracts/AssetPool.sol
pdyraga: What if we say "... plus 2 days to make a decision"? 

It's quite an arbitrary value, to be honest. I assumed 2 days are enough in most cases to be notified and make a decision, maybe it is not enough.

dimpar: Sure, fair enough.

pdyraga: Actually, worth mentioning that every `finalize*` governance function is `onlyOwner`, so if the governance wants to give some extra time for underwriters they can do that at any moment. Those two days are an absolute minimum. I'll clarify it in the docs.

pdyraga: Done as a part of https://github.com/keep-network/coverage-pools/pull/105/commits/f6c542d35278810ac95b149495b3edb6858cc149.

contracts/CoveragePool.sol
pdyraga: Good point, done in https://github.com/keep-network/coverage-pools/pull/105/commits/137849942d8ce99c5ffcece349aaec70bf039899.

pdyraga:
Okay, I am happy with what is here now so I am going to undraft this PR.

contracts/AssetPool.sol
```diff
@@ -72,13 +77,37 @@ contract AssetPool is Ownable, IAssetPool {
...
+        uint256 timestamp
+    );
+    event WithdrawalDelayUpdated(uint256 withdrawalDelay);
+    event WithdrawalTimeoutUpdateStarted(
+        uint256 withdrawalTimeout,
+        uint256 timestamp
+    );
+    event WithdrawalTimeoutUpdated(uint256 withdrawalTimeout);
+
+    modifier onlyAfterWithdrawalGovernanceDelay(
```
lukasz-zimnoch, line 99: Non-blocking (#https://github.com/keep-network/coverage-pools/issues/82): We have all other modifiers documented. Can we document this one as well?

contracts/AssetPool.sol
```diff
@@ -296,6 +325,113 @@ contract AssetPool is Ownable, IAssetPool {
...
+        emit WithdrawalTimeoutUpdated(withdrawalTimeout);
+        newWithdrawalTimeout = 0;
+        withdrawalTimeoutChangeInitiated = 0;
+    }
+
+    /// @notice Returns the remaining time that has to pass before the contract
+    ///         owner will be able to finalize withdrawal delay update.
+    ///         Bear in mind the contract owner may decide to wait longer and
+    ///         this value is just an absolute minimum.
+    function getRemainingWithdrawalDelayUpdateTime()
```
lukasz-zimnoch, line 398: Non-blocking (https://github.com/keep-network/coverage-pools/issues/82): Can we document the returned value using `@return`?

contracts/AssetPool.sol
```diff
@@ -296,6 +325,113 @@ contract AssetPool is Ownable, IAssetPool {
...
+                withdrawalDelayChangeInitiated,
+                withdrawalGovernanceDelay()
+            );
+    }
+
+    /// @notice Returns the remaining time that has to pass before the contract
+    ///         owner will be able to finalize withdrawal timeout update.
+    ///         Bear in mind the contract owner may decide to wait longer and
+    ///         this value is just an absolute minimum.
+    function getRemainingWithdrawalTimeoutUpdateTime()
```
lukasz-zimnoch, line 414: Non-blocking (https://github.com/keep-network/coverage-pools/issues/82): Can we document the returned value using `@return`?

contracts/AssetPool.sol
```diff
@@ -296,6 +325,113 @@ contract AssetPool is Ownable, IAssetPool {
...
+                withdrawalGovernanceDelay()
+            );
+    }
+
+    /// @notice The time it takes to initiate and complete the withdrawal from
+    ///         the pool plus 2 days to make a decision. This governance delay
+    ///         should be used for all changes directly affecting underwriter
+    ///         positions. This time is a minimum and the governance may choose
+    ///         to wait longer before finalizing the update.
+    function withdrawalGovernanceDelay() public view returns (uint256) {
```
lukasz-zimnoch, line 431: Non-blocking (https://github.com/keep-network/coverage-pools/issues/82): Can we document the returned value using `@return`?

contracts/AssetPool.sol
```diff
@@ -296,6 +325,113 @@ contract AssetPool is Ownable, IAssetPool {
...
+        return withdrawalDelay.add(withdrawalTimeout).add(2 days);
```
lukasz-zimnoch, line 432: Maybe it would be good to extract the `2 days` into a constant? Also, is there a particular reason explaining why is it exactly 2 days?

lukasz-zimnoch: Ok I see the answer for the second question: https://github.com/keep-network/coverage-pools/pull/105#discussion_r653415075

dimpar: Piotr is experiencing déjà vu :) https://github.com/keep-network/coverage-pools/pull/105#discussion_r653386129

contracts/AssetPool.sol
pdyraga: https://github.com/keep-network/coverage-pools/pull/105/commits/c0764c44ac0bc0884c60130ce06bce5fefa46cb8

contracts/AssetPool.sol
pdyraga: https://github.com/keep-network/coverage-pools/pull/105/commits/c0764c44ac0bc0884c60130ce06bce5fefa46cb8

contracts/AssetPool.sol
pdyraga: https://github.com/keep-network/coverage-pools/pull/105/commits/c0764c44ac0bc0884c60130ce06bce5fefa46cb8

contracts/AssetPool.sol
pdyraga: https://github.com/keep-network/coverage-pools/pull/105/commits/c0764c44ac0bc0884c60130ce06bce5fefa46cb8

contracts/AssetPool.sol
pdyraga: I am not sure about extracting 2 days into a constant, this is the only place it is used and I think in the context of this function it is best explained what is it for... WDYT?

lukasz-zimnoch: Sure. Let's leave as is.

lukasz-zimnoch (approved):

Issue #98, opened 2021-06-11
Standarize Solidity version used


pdyraga:
Some of the contracts use 0.8, some others 0.7. We need to standardize the Solidity version referenced, probably to 0.7.

ggrieco-tob:
Related with this, I believe the pragma used by several contracts (`pragma solidity <0.9.0;`) is too broad and can induce security or correctness issues. Since such pragma allows 0.7.x and 0.8.x, the developers can overlook certain features if they have to downgrade the compiler versions.

**Exploit scenario**

If a contract is compiled using 0.8.x, solc will include mandatory integer overflow checks. As expected, contract developers are tempted to remove the use of SafeMath, since it is redundant and waste gas. However, if the optimized contract is later compiled with solc 0.7.x, the integer overflow checks will be silently removed.

**Recommendation**

Select either 0.7.x or 0.8.x solc versions to compile contracts and include a suitable pragma for it.

pdyraga:
Confirmed with Trail of Bits: 

> In terms of solc versions, 0.8.x is fine, as long it's not the latest release (which sometimes contains latent issues). Right now, latest release is 0.8.5, so 0.8.4 is fine, in particular since the fixed a high-severity issue related with the decoder.

I recommend we stick to 0.8.4, and review the contracts so they use 0.8.x features, especially review all `SafeMath` usages.

PR #107, opened 2021-06-17
Standardise solidity version to 0.8.4

dimpar:
Closes https://github.com/keep-network/coverage-pools/issues/98

For Solidity upgrade to 0.8.4 this PR also upgrades OZ to 4.1.0 to align with Solidity compiler. Some of the compilation errors had to be changed as well.

### DIFF
--- a/contracts/AssetPool.sol
+++ b/contracts/AssetPool.sol
@@ -175,7 +175,7 @@ contract AssetPool is Ownable, IAssetPool {
             "Underwriter token amount must be greater than 0"
         );
 
-        pending = pending + covAmount;
+        pending += covAmount;
         pendingWithdrawal[msg.sender] = pending;
         /* solhint-disable not-rely-on-time */
         withdrawalInitiatedTimestamp[msg.sender] = block.timestamp;
--- a/contracts/AssetPool.sol
+++ b/contracts/AssetPool.sol
@@ -16,7 +16,6 @@ pragma solidity 0.8.4;
 
 import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
-import "@openzeppelin/contracts/utils/math/SafeMath.sol";
 import "@openzeppelin/contracts/access/Ownable.sol";
 
 import "./interfaces/IAssetPool.sol";
@@ -35,7 +34,6 @@ import "./GovernanceUtils.sol";
 contract AssetPool is Ownable, IAssetPool {
     using SafeERC20 for IERC20;
     using SafeERC20 for UnderwriterToken;
-    using SafeMath for uint256;
 
     IERC20 public collateralToken;
     UnderwriterToken public underwriterToken;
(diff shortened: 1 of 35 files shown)


contracts/AssetPool.sol
```diff
@@ -12,11 +12,11 @@
...
 // SPDX-License-Identifier: MIT
 
-pragma solidity <0.9.0;
+pragma solidity 0.8.4;
 
 import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
-import "@openzeppelin/contracts/token/ERC20/SafeERC20.sol";
-import "@openzeppelin/contracts/math/SafeMath.sol";
+import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import "@openzeppelin/contracts/utils/math/SafeMath.sol";
```
pdyraga, line 19: Do we still need it?

dimpar: Nope, [Solidity docs](https://docs.soliditylang.org/en/latest/080-breaking-changes.html#how-to-update-your-code) tells to replace it. Done in c154802
However, one file is still untouched `ERC20WithPermit.sol` which includes custom error messages for `.sub()`. Do we want to refactor that as well along with the tests?

contracts/ERC20WithPermit.sol
```diff
@@ -12,9 +12,9 @@
 
 // SPDX-License-Identifier: MIT
 
-pragma solidity <0.8.0;
+pragma solidity 0.8.4;
 
-import "@openzeppelin/contracts/math/SafeMath.sol";
+import "@openzeppelin/contracts/utils/math/SafeMath.sol";
```
tomaszslabon, line 17: I `SafeMath` still needed in `ERC20WithPermit.sol`?
I see that `add()` is still used, but it could probably be replaced with `+`.

contracts/AssetPool.sol
pdyraga: https://github.com/keep-network/coverage-pools/pull/107/commits/eb8d0546c3bf58d1b065baee638d7f40d830b414 😜 

We do want to keep the same revert messages. I started looking into it, then it escalated... and I figured I'll just commit and leave the review of this change to you.

dimpar:
@pdyraga all the comments related to operators are addressed here: 79af193

> Can we remove uint256 from this conversion?

I've tried that and all the tests passed, but... according to the docs: 

> `address(uint)` and `uint(address)` : converting both type-category and width. Replace this by `address(uint160(uint))` and `uint(uint160(address))` respectively.

 I'm just not sure if there can be some weird corner case where having just `tbtcDepositToken.exists(uint160(depositAddress))` breaks things.

pdyraga:
> > Can we remove uint256 from this conversion?
> 
> I've tried that and all the tests passed, but... according to the docs:
> 
> > `address(uint)` and `uint(address)` : converting both type-category and width. Replace this by `address(uint160(uint))` and `uint(uint160(address))` respectively.
> 
> I'm just not sure if there can be some weird corner case where having just `tbtcDepositToken.exists(uint160(depositAddress))` breaks things.

I think it is safe, it's just a matter of whether `uint160 -> uint256` conversion happens implicitly or explicitly. We properly updated `uint(address)` to `uint(uint160(address))` per Solidity 8 changelog but the additional `uint256` should have nothing to do about Solidity 8. That's fine, it's just a detail, we can leave it as-is.

pdyraga (approved):

Issue #99, opened 2021-06-14
Auctions for deposits with FRAUD_LIQUIDATION_IN_PROGRESS status not allowed


tomaszslabon:
Auctions can only be opened for deposits with status `DEPOSIT_LIQUIDATION_IN_PROGRESS_STATE`.
Perhaps auction for deposits with `FRAUD_LIQUIDATION_IN_PROGRESS` status should also be allowed.

Note that in the `purchaseSignerBondsAtAuction` function in tBTC project we have the following check:
```
require(_d.inSignerLiquidation(), "No active auction");```
```
which is:
```
function inSignerLiquidation(DepositUtils.Deposit storage _d) external view returns (bool) {
    return (
        _d.currentState == uint8(States.LIQUIDATION_IN_PROGRESS)
     || _d.currentState == uint8(States.FRAUD_LIQUIDATION_IN_PROGRESS)
    );
}
```

In other words, you can liquidate deposits with `FRAUD_LIQUIDATION_IN_PROGRESS` outside Coverage Pools.

PR #84, opened 2021-06-03
[System tests] deposit liquidated outside Coverage Pools

tomaszslabon:
This pull request:
- modifies `RiskManagerV1` so that the check of deposit's state is performed explicitly when auctions fully filled
-  adds a system test for a scenario where the deposit is liquidated outside Coverage Pools.
  There are two test cases:
  - auction is partially filled (deposit status is checked by Coverage Pools in `onAuctionPartiallyFilled`)
  - auction is fully filled (deposit status is checked by Coverage Pools in `onAuctionPartiallyFilled`)
- removes an integration test for the same scenario (scenarios involving handling tBTC deposits are better tested using system tests).

Refs https://github.com/keep-network/coverage-pools/issues/83

PR #109, opened 2021-06-18
Allow auctions for fraud liquidations

lukasz-zimnoch:
Closes: https://github.com/keep-network/coverage-pools/issues/99

The risk manager should accept all deposits which are in liquidation due to fraud.

### DIFF
--- a/contracts/RiskManagerV1.sol
+++ b/contracts/RiskManagerV1.sol
@@ -417,7 +417,7 @@ contract RiskManagerV1 is IRiskManager, Auctioneer, Ownable {
         IDeposit deposit = IDeposit(auctionToDeposit[address(auction)]);
         // Make sure the deposit was not liquidated outside of Coverage Pool
         require(
-            deposit.currentState() == DEPOSIT_LIQUIDATION_IN_PROGRESS_STATE,
+            isDepositLiquidationInProgress(deposit),
             "Deposit liquidation is not in progress"
         );
 
--- a/contracts/RiskManagerV1.sol
+++ b/contracts/RiskManagerV1.sol
@@ -70,6 +70,7 @@ contract RiskManagerV1 is IRiskManager, Auctioneer, Ownable {
 
     uint256 public constant GOVERNANCE_TIME_DELAY = 12 hours;
 
+    uint256 public constant DEPOSIT_FRAUD_LIQUIDATION_IN_PROGRESS_STATE = 9;
     uint256 public constant DEPOSIT_LIQUIDATION_IN_PROGRESS_STATE = 10;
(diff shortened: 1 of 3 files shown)


contracts/RiskManagerV1.sol
```diff
@@ -454,7 +455,7 @@ contract RiskManagerV1 is IRiskManager, Auctioneer, Ownable {
         IDeposit deposit = IDeposit(auctionToDeposit[address(auction)]);
         // Make sure the deposit was not liquidated outside of Coverage Pool
         require(
-            deposit.currentState() == DEPOSIT_LIQUIDATION_IN_PROGRESS_STATE,
+            isDepositLiquidationInProgress(deposit),
```
pdyraga, line 458: After the change from #84, we need to add this check to `onAuctionFullyFilled`.

pdyraga (commented):
One comment we need to address in this PR. Otherwise looks good.

contracts/RiskManagerV1.sol
lukasz-zimnoch: Sure. Here it is: https://github.com/keep-network/coverage-pools/pull/109/commits/9478d91c786edfa234ef548783220003648c8097

pdyraga:
@lukasz-zimnoch Can you please resolve conflicts?

lukasz-zimnoch:
> @lukasz-zimnoch Can you please resolve conflicts?

Done!

contracts/test/DepositStub.sol
```diff
@@ -85,4 +85,8 @@ contract DepositStub is IDeposit {
     function withdrawableAmount() external view override returns (uint256) {
         return address(this).balance;
     }
+
+    function notifyFraud() external {
```
pdyraga, line 89: This is not a part of a real Deposit contract public API? If so, we should add a note about it, just like we did with `setAuctionValue`.

pdyraga: Also, linter is complaining about the function order, can we fix it?

lukasz-zimnoch: It's under:
```
    ///
    /// Not in tBTC deposit interface, added just for tests.
    ///
```

Regarding the order, I'll change.

pdyraga: No, this comment is now just for that one function, because `withdrawableAmount` is a part of the public interface and it is located below. I had to chang the approach because linter was annoying in PRs :/

lukasz-zimnoch: Ok, done: https://github.com/keep-network/coverage-pools/pull/109/commits/363e308394a507077a3c4aeffaba22c7cfe6798b

pdyraga (approved):

Issue #80, opened 2021-06-03
`notifyLiquidation` and `notifyLiquidated` incentives


pdyraga:
Right now, there are no incentives for anyone to notify the risk manager. The only case when someone would do that non-voluntarily is when they would call`notifyLiquidation` to open an auction and then bid in that auction immediately. 

Ideally, the incentivization mechanism should not grant more rewards from a cookie jar but rather grant assets it "has", which in the case of a single-asset coverage pool is probably COV token. Minting COV without depositing underlying, from the underwriter's perspective, is similar to a coverage claim - the value of their COV is getting lower.

It feels that a fixed governable reward could work better, considering how poorly %-based worked for a few tBTC v1 mechanisms. On the other hand, COV tokens have shifting value relative to the underlying, so it's hard for governance to choose a fee that makes sense over time, in which case maybe governance should instead choose a %.

pdyraga:
In all cases I can think of, a fixed amount looks better to me:
- If the pool TVL grows fast, % is more dangerous and potentially needs to be updated several times as the pool grows.
- The value of COV is changing, true. But % approach does not help, unless we calculate it as % of the underlying deposited in the pool which makes the mechanism complex, and finding the right value by the governance could be hard as well.

pdyraga:
Tagging @mhluongo for a discussion.

lukasz-zimnoch:
@pdyraga I opened a draft with a fixed amount: https://github.com/keep-network/coverage-pools/pull/111. I'd appreciate some opinions about the approach before I continue with the rest.

PR #111, opened 2021-06-18
Notifier reward

lukasz-zimnoch:
Closes: https://github.com/keep-network/coverage-pools/issues/80

This pull request introduces incentives for `notifyLiquidation` and `notifyLiquidated` calls. A notifier performing one of those actions should be rewarded by obtaining some of the underwriter tokens representing the asset pool ownership. The exact amount of granted tokens is set and updated by the governance.


contracts/RiskManagerV1.sol
```diff
@@ -98,6 +98,12 @@ contract RiskManagerV1 is IRiskManager, Auctioneer, Ownable {
     ISignerBondsSwapStrategy public newSignerBondsSwapStrategy;
     uint256 public signerBondsSwapStrategyInitiated;
 
+    // Amount of underwriter tokens granted to the notifier as reward for
+    // `notifyLiquidation` and `notifyLiquidated` calls.
+    uint256 public notifierReward;
+    uint256 public newNotifierReward;
+    uint256 public notifierRewardChangeInitiated;
```
pdyraga, line 105: Can we put them after `auctionLengthChangeInitiated` so that all `uint256` fields are grouped together?

pdyraga, line 105: I would consider having two rewards: one for `notifyLiquidating` and another for `notifyLiquidated` in case it is needed.

pdyraga: I would also consider kicking the can on whether we go for % or with a fixed amount in the following way:
- if the fixed COV amount is specified, we go with the fixed amount,
- otherwise, we go with % of the COV supply.

It should be possible to set the reward to be << 1 %.

I would also consider putting all this logic in a library, similarly how tBTC `Deposit` is split into libraries. This is too much for one contract. And I am thinking about extracting some more of the code from this contract into a library.

lukasz-zimnoch: Just to summarize, we want to have the following:

- `liquidationNotifierRewardAmount`
- `liquidationNotifierRewardPercentage`
- `liquidatedNotifierRewardAmount`
- `liquidatedNotifierRewardPercentage`

All parameters should be governable. `*Amount` parameters should be settable to `0`. `*Percentage` params should be settable to value less than `1%` (is 0.01% at min enough?). If `*Amount` param is `0` we use `*Percentage` of the total COV supply. All this stuff should live in a separate lib. Am I right?

pdyraga: Yes. 🧔 

And to clarify:

> (is 0.01% at min enough?)

Let's use 1e18 precision for %, just like we do for cov pool on auction.

lukasz-zimnoch: I think it will be no longer applicable due to https://github.com/keep-network/coverage-pools/pull/111#discussion_r656862995

lukasz-zimnoch: Ok extracted the library. Please see https://github.com/keep-network/coverage-pools/pull/111/commits/b3d880775271569a41219707797f574e7ad95486. There are still some governance methods missing in the risk manager but it shows the general idea. It also doesn't touch the tests yet. More to come.

contracts/AssetPool.sol
```diff
@@ -391,6 +391,21 @@ contract AssetPool is Ownable, IAssetPool {
         withdrawalTimeoutChangeInitiated = 0;
     }
 
+    /// @notice Grants pool shares by minting a given amount of the underwriter
+    ///         tokens for the recipient address. In result, the recipient
+    ///         obtains part of the pool ownership without depositing any
+    ///         collateral tokens.
```
pdyraga, line 397: Worth adding a note in what circumstances this function could be used (rewards for various contract state notifiers).

contracts/CoveragePool.sol
```diff
@@ -188,6 +188,20 @@ contract CoveragePool is Ownable {
         assetPool.claim(recipient, amountToSeize(portionToSeize));
     }
 
+    /// @notice Grants asset pool shares by minting a given amount of the
+    ///         underwriter tokens for the recipient address. In result, the
+    ///         recipient obtains part of the pool ownership without depositing
+    ///         any collateral tokens.
```
pdyraga, line 194: Here as well, let's add a note in what circumstances shares are granted.

contracts/NotifierRewards.sol
```diff
@@ -0,0 +1,263 @@
...
+
+import "./CoveragePoolConstants.sol";
+
+/// @title NotifierRewards
+/// @notice Contains logic responsible for calculating notifier rewards for
+///         both deposit liquidation start and deposit liquidated events.
+///         All parameters can be updated using a two-phase process.
+/// @dev The client contract should take care of authorizations or governance
+///      delays according to their needs.
+library NotifierRewards {
```
pdyraga, line 25: What do you think about naming it `RiskManagerV1Rewards`? It is not generic library that can be used for any risk manager but something specific to tBTC v1 because of the assumption there is `notifyLiquidating` and `notifyLiquidated`. If it is possible and the linter is not complaining, we could even put this code in `RiskManagerV1.sol` file. This way, there is only one tbtc v1 specific file and we will not be tempted to introduce a separate directory per risk manager.

lukasz-zimnoch: Done in https://github.com/keep-network/coverage-pools/pull/111/commits/642d8147381b16b9b91d1d2cc98b580e4159488e

contracts/RiskManagerV1.sol
```diff
@@ -461,6 +652,13 @@ contract RiskManagerV1 is IRiskManager, Auctioneer, Ownable {
         );
     }
 
+    /// @return Total supply of the COV token.
+    function covTotalSupply() internal view returns (uint256) {
+        UnderwriterToken underwriterToken =
+            AssetPool(coveragePool.assetPool()).underwriterToken();
+        return underwriterToken.totalSupply();
+    }
```
pdyraga, line 657: It feels that underwriter token calculations are not something the risk manager should care about. We could introduce function similar to `amountToSeize` in `CoveragePool` to compute the amount of underwriter tokens one should receive based on the portion of the pool to grant. 

We do have `amountToSeize(uint256 portionToSeize)` as an amount of collateral tokens to seize and we could have next to it `amountToGrant(uint256 portionToGrant)` as an amount of underwriter tokens to grant. What do you think?

pdyraga: Ah, I see it now. You need the amount because the reward is either fixed amount or % and we need to calculate this %.

contracts/RiskManagerV1.sol
```diff
@@ -122,6 +127,12 @@ contract RiskManagerV1 is IRiskManager, Auctioneer, Ownable {
         address indexed signerBondsSwapStrategy
     );
 
+    event NotifierRewardUpdateStarted(
+        uint256 notifierReward,
+        uint256 timestamp
+    );
+    event NotifierRewardUpdated(uint256 notifierReward);
```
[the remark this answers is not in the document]
lukasz-zimnoch: Done in https://github.com/keep-network/coverage-pools/pull/111/commits/e9c6b6cc0fbd4066c472bc93916e8d954c2d07c9

contracts/RiskManagerV1.sol
```diff
@@ -472,3 +661,249 @@ contract RiskManagerV1 is IRiskManager, Auctioneer, Ownable {
...
+        uint256 liquidatedNotifierRewardPercentage
+    );
+
+    /// @notice Begins the liquidation notifier reward amount update process.
+    /// @param _newLiquidationNotifierRewardAmount New liquidation notifier
+    ///        reward amount.
+    function beginLiquidationNotifierRewardAmountUpdate(
+        Storage storage self,
+        uint256 _newLiquidationNotifierRewardAmount
+    ) external {
```
pdyraga, line 736: We can probably keep them all `internal`, the optimizer should copy the internal function code.

lukasz-zimnoch: But that will cause a growth of the `RiskManagerV1` bytecode size. Do we want that for sure?

contracts/RiskManagerV1.sol
pdyraga: On the other hand, given that all functions are internal to the reward manager, wouldn't it be more elegant:
```
    function getLiquidationNotifierReward(
        Storage storage self,
        CoveragePool coveragePool
    ) internal view returns (uint256) {
        if (self.liquidationNotifierRewardAmount > 0) {
            return self.liquidationNotifierRewardAmount;
        }

        return coveragePool.amountToGrant(self.liquidationNotifierRewardPercentage);
    }
```
?

lukasz-zimnoch: Sounds good. See https://github.com/keep-network/coverage-pools/pull/111/commits/3343bab820f506e1d1df2464cd601e19995dd516. I just named it `covAmountToGrant` to not confuse with `amountToSeize` which is about collateral tokens.

contracts/RiskManagerV1.sol
```diff
@@ -298,6 +314,110 @@ contract RiskManagerV1 is IRiskManager, Auctioneer, Ownable {
         auctionLengthChangeInitiated = 0;
     }
 
+    /// @notice Begins the liquidation notifier reward amount update process.
+    /// @dev Can be called only by the contract owner.
+    /// @param _newLiquidationNotifierRewardAmount New liquidation notifier
+    ///        reward amount.
+    function beginLiquidationNotifierRewardAmountUpdate(
```
pdyraga, line 321: I'd skip the *Amount* part from function names and parameters. Meaning is the same and they can be shorter.

contracts/RiskManagerV1.sol
```diff
@@ -472,3 +752,191 @@ contract RiskManagerV1 is IRiskManager, Auctioneer, Ownable {
...
+
+    /// @notice Finalizes the liquidated notifier reward percentage update process.
+    function finalizeLiquidatedNotifierRewardPercentageUpdate(
+        Storage storage self
+    ) internal {
+        self.liquidatedNotifierRewardPercentage = self
+            .newLiquidatedNotifierRewardPercentage;
+        self.newLiquidatedNotifierRewardPercentage = 0;
+        self.liquidatedNotifierRewardPercentageChangeInitiated = 0;
+    }
```
pdyraga, line 899: I am thinking we could combine liquidating and liquidated rewards update into two functions given that it is always either amount or percentage. Small detail though, I am good with having it as-is but figured I'll leave the note just in case it helps with some change in the future.

contracts/RiskManagerV1.sol
lukasz-zimnoch: Done.

contracts/CoveragePool.sol
```diff
@@ -216,4 +230,19 @@ contract CoveragePool is Ownable {
             (collateralToken.balanceOf(address(assetPool)) * portionToSeize) /
             CoveragePoolConstants.FLOATING_POINT_DIVISOR;
     }
+
+    /// @notice Calculates the amount of COV tokens for a grant.
+    /// @param portionToGrant Portion to grant in the range [0, 1] multiplied
+    ///        by FLOATING_POINT_DIVISOR.
```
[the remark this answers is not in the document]
lukasz-zimnoch: See https://github.com/keep-network/coverage-pools/pull/111/commits/b90c313057259b83d47f3c0b60072bf4d6881698

contracts/test/CoveragePoolStub.sol
```diff
@@ -1,10 +1,35 @@
...
+        external
+        view
+        returns (uint256)
+    {
+        return
+            (portionToGrant * covTotalSupply) /
+            CoveragePoolConstants.FLOATING_POINT_DIVISOR;
+    }
+
+    function setCovTotalSupply(uint256 _covTotalSupply) external {
```
pdyraga, line 32: Is it used anywhere?

contracts/test/CoveragePoolStub.sol
```diff
@@ -1,10 +1,35 @@
...
         emit FundsSeized(recipient, portionToSeize);
     }
+
+    function grantAssetPoolShares(address recipient, uint256 covAmount)
+        external
+    {
+        emit AssetPoolSharesGranted(recipient, covAmount);
+    }
+
+    function covAmountToGrant(uint256 portionToGrant)
```
pdyraga, line 22: Do we have to implement this function here?

lukasz-zimnoch: Yep, some tests are using the coverage pool stub and look for that method.

contracts/test/CoveragePoolStub.sol
lukasz-zimnoch: Nope, I added it just in case but was not needed eventually. I'll remove.

lukasz-zimnoch: See https://github.com/keep-network/coverage-pools/pull/111/commits/2d11ae71aef5b39ec9329a3ca96e82b002284be4

test/RiskManagerV1.test.js
```diff
@@ -279,6 +320,12 @@ describe("RiskManagerV1", () => {
                   .withArgs(depositStub.address, notifier.address)
               })
 
+              it("should reward the notifier with asset pool shares", async () => {
+                await expect(notifyLiquidationTx)
+                  .to.emit(coveragePoolStub, "AssetPoolSharesGranted")
+                  .withArgs(notifier.address, liquidationNotifierRewardAmount)
+              })
```
pdyraga, line 327: Wouldn't it be better to keep only one instance of this test under
```
context("when deposit is in liquidation state", () => {
  context("when deposit is at the bond auction threshold", () => {
```
? The logic of withdrawing does not change depending on the surplus etc.

lukasz-zimnoch: I wish we can do that but unfortunatelly we need `notifyLiquidationTx` which is not available here.

contracts/AssetPool.sol
lukasz-zimnoch: See https://github.com/keep-network/coverage-pools/pull/111/commits/5b3b64029f99da1394f103cc2c7db4f05cdc7966

contracts/CoveragePool.sol
lukasz-zimnoch: See https://github.com/keep-network/coverage-pools/pull/111/commits/5b3b64029f99da1394f103cc2c7db4f05cdc7966

contracts/test/CoveragePoolStub.sol
pdyraga: Hmm.... I commented out this function and run all tests (unit + system) and all is green. If tests use this function then something is probably wrong because it returns 0 now (`covTotalSupply` is zero).

test/system/notifier-rewards.test.js
```diff
@@ -0,0 +1,285 @@
...
+        await riskManagerV1.finalizeLiquidationNotifierRewardPercentageUpdate()
+
+        await riskManagerV1
+          .connect(notifier)
+          .notifyLiquidation(depositStub.address)
+      })
+
+      it("should be rewarded with percentage-based amount of asset pool shares", async () => {
+        expect(await underwriterToken.balanceOf(notifier.address)).to.be.equal(
+          covTotalSupply.div(50) // liquidationNotifierRewardPercentage is 2%
```
pdyraga, line 185: There are no deposits so this is... 0? 

I would rather used a fixed value here instead of computing it so that we can avoid this kind of problems.

pdyraga:
Two last comments on my side @lukasz-zimnoch. Let's also update `docs/design.adoc` but we can do it in a separate PR if you prefer it.

docs/design.adoc
```diff
@@ -314,3 +314,31 @@ liquidating deposits still attractive for arbitraging bots.
...
+* For `notifyLiquidated` rewards:
+** `liquidatedNotifierRewardAmount`: determines the fixed amount of COV tokens
+which is granted as reward for reporting about a deposit being liquidated
+outside of the coverage pool.
+** `liquidatedNotifierRewardPercentage`: determines the percentage of the
+total COV supply which is granted as reward for reporting about a deposit
+being liquidated outside of the coverage pool.
+
+If both fixed amount and percentage values are set and bigger than zero,
+the fixed amount reward takes precedence.
```
pdyraga, line 344: We also need to update the Governance section FWIW.

docs/design.adoc
```diff
@@ -314,3 +314,31 @@ liquidating deposits still attractive for arbitraging bots.
 |12 hours
 |Governance can chose which strategy should be used by tBTC v1 risk manager for
 depositing ETH signer bonds purchased from tBTC to the asset pool.
+|===
+
+== Notifier rewards
```
pdyraga, line 319: Shouldn't it live under tBTC v1 Risk Manager?

test/system/notifier-rewards.test.js
lukasz-zimnoch: Nope, the COV supply is minted at the beginning. I agree fixed amount will be better. See https://github.com/keep-network/coverage-pools/pull/111/commits/d1e910303abbb5ac89fe32c0ab6d267a08aa99ce

contracts/test/CoveragePoolStub.sol
lukasz-zimnoch: Ah I see now, it was needed before but it's no longer the case. See https://github.com/keep-network/coverage-pools/pull/111/commits/e7a6f40c79bfcfd0bd82e56ef102f6a282e78380

docs/design.adoc
lukasz-zimnoch: Addressed https://github.com/keep-network/coverage-pools/pull/111/commits/ef8b524c44f1ebac4bd0e5851ecb3d5ae5a2bef8

docs/design.adoc
lukasz-zimnoch: See https://github.com/keep-network/coverage-pools/pull/111/commits/ef8b524c44f1ebac4bd0e5851ecb3d5ae5a2bef8

pdyraga (approved):

PR #116, opened 2021-06-23
Increased withdrawal delay to 21 days

### LINKED ISSUE #89: Simplify withdrawal mechanism?

pdyraga:
Closes #89 

We were looking at historical liquidations and 14 days withdrawal delay
could not be enough in all cases. Of course, those are just historical
data and there is no guarantee future will look similar but if something
already happened there is a chance it will happen again and we should be
prepared.

All liquidations we had on mainnet can be reduced to three cases:
- liquidations around ~Nov 5th,
- liquidations around ~Nov 17th,
- liquidations around ~Dec 26th.

If I were coverage pool underwriter having a strategy to withdraw all my
funds when it starts getting risky setting the threshold at 135%
collateralization level, two week withdrawal period would not stop me
from escaping the pool for Nov 17th liquidation. We can not stop people
from withdrawing but the whole point of a coverage pool is to be with
the pool when it is sunny outside, everyone is happy, and earning
rewards, and be with the pool when the storm is coming so that there is
that last-chance insurance. We do not want to make it easy for
underwriters trying to escape when they see a small cloud on the sky
and being interested in a free ride on everyone else's shoulders. 21 days
withdrawal delay feels like a good tradeoff.

aaa1

aaa2

aaa3

### DIFF
--- a/contracts/AssetPool.sol
+++ b/contracts/AssetPool.sol
@@ -47,7 +47,7 @@ contract AssetPool is Ownable, IAssetPool {
     // initiating and completing the withdrawal. During that time, underwriter
     // is still earning rewards and their share of the pool is still a subject
     // of a possible coverage claim.
-    uint256 public withdrawalDelay = 14 days;
+    uint256 public withdrawalDelay = 21 days;
     uint256 public newWithdrawalDelay;
     uint256 public withdrawalDelayChangeInitiated;
     // The time the underwriter has after the withdrawal delay passed to
--- a/test/AssetPool.test.js
+++ b/test/AssetPool.test.js
@@ -513,7 +513,7 @@ describe("AssetPool", () => {
           .approve(assetPool.address, amount)
         await assetPool.connect(underwriter1).initiateWithdrawal(amount)
 
-        await increaseTime(14 * 86400 - 1) // 14 days - 1 sec
+        await increaseTime(21 * 86400 - 1) // 21 days - 1 sec
       })
 
       it("should revert", async () => {
@@ -533,7 +533,7 @@ describe("AssetPool", () => {
           .connect(underwriter1)
           .approve(assetPool.address, amount)
         await assetPool.connect(underwriter1).initiateWithdrawal(amount)
-        await increaseTime((14 + 2) * 86400) // 14 + 2 days
+        await increaseTime((21 + 2) * 86400) // 21 + 2 days
       })
 
       it("should revert", async () => {
@@ -554,7 +554,7 @@ describe("AssetPool", () => {
           .connect(underwriter1)
           .approve(assetPool.address, amount)
         await assetPool.connect(underwriter1).initiateWithdrawal(amount)
-        await increaseTime(14 * 86400) // 14 days
+        await increaseTime(21 * 86400) // 21 days
         const tx = await assetPool.completeWithdrawal(underwriter1.address)
 
         expect(tx)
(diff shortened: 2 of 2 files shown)


dimpar:
In `assetPool.test.js` line 434 we still have `14` days. What do you think about extracting it to test level constant?

pdyraga:
> In `assetPool.test.js` line 434 we still have `14` days. What do you think about extracting it to test level constant?

Yeah, good catch and a very good idea. Done in the last commit.

dimpar (approved):
lgtm

tomaszslabon (approved):

Issue #112, opened 2021-06-21
Get a total pool share value in one function call


pdyraga:
To get a total pool share based on one's COV, they need to get the total collateral balance of `AssetPool` and learn about `RewardsPool.earned()`. We could make it easier for dApp developers to get this information in one call from `AssetPool` as it is referencing its `RewardsPool`.

PR #121, opened 2021-06-25
Finishing touches: get pool's TVL and claim history

pdyraga:
Closes: #112 

Two changes here:

- Emit `CovrageClaimed` event from `AssetPool.claim`

  We were not emitting an event from the asset pool when a portion of tokens
was claimed as coverage. Having this event in the pool can help to
calculate historical APY without having to look at risk manager
implementation. This is handy because we may have multiple
implementations of a risk manager in the future.

- Added `AssetPool.totalValue` function

  The function returns the current collateral token balance of the asset
pool plus the reward amount (in collateral token) earned by the asset
pool and not yet withdrawn to the asset pool.

  This function will be helpful for dApp to compute the current TVL of the
asset pool without having to query asset pool and rewards pool
separately.


test/AssetPool.test.js
```diff
@@ -910,6 +922,51 @@ describe("AssetPool", () => {
     })
   })
 
+  describe("totalValue", () => {
+    context("when there is nothing in the pool", () => {
+      it("should return zero", async () => {
+        expect(await assetPool.totalValue()).to.equal(to1e18(0))
```
dimpar, line 928: I think we can assert `to.equal(0)`

dimpar:
Seems like emitting event bumps up gas cost.
Rest of the code looks good.

test/AssetPool.test.js
pdyraga: https://github.com/keep-network/coverage-pools/pull/121/commits/9310e0287d0390342f90f04143ae6a3ee215767c

dimpar (approved):

PR #128, opened 2021-06-28
Hardhat coverage plugin

tomaszslabon:
This PR:
- adds the [Hardhat coverage plugin](https://hardhat.org/plugins/solidity-coverage.html)
- adds rules for ignoring files generated by the coverage plugin in `.gitignore`
- adds yarn script for running coverage tests. It can be run with:
```
yarn test:unit:coverage
```

NOTE: Only the coverage of unit tests are taken for coverage report.

The `solidity-coverage` plugin does not support the `hardhat_reset` functionality that is used in the system tests: https://github.com/sc-forks/solidity-coverage/issues/574.
Until the issue in `solidity-coverage`is resolved, it will not be possible to generate coverage reports with our system tests, unless a workaround is found.


pdyraga:
Early question without looking at the code: will the coverage report include both unit and system tests? This would be ideal.

tomaszslabon:
> Early question without looking at the code: will the coverage report include both unit and system tests? This would be ideal.

@pdyraga 
Command 
```
FORKING_URL=https://eth-mainnet.alchemyapi.io/v2/<your_alchemy_key> yarn coverage
```
runs both unit tests and system tests.
The system tests are run (I checked it with logs), but I've noticed some problem: they don't seem to affect the coverage results -  for example the coverage for `SignerBondsUniswapV2.sol` is 0%, but there is a system test for this contract and it runs.

I decided to comment out all unit tests, run it again, and the results show coverage for **interfaces** (like `IAssetPool.sol`, `IAuction.sol`, `IRiskManagerV1.sol`) instead of **concrete instances** (`AssetPool.sol`, `Auction.sol`, `RiskManagerV1.sol`). I'm trying to find why this is happening.

dimpar:
I see that `package-lock.json` was generated. We already have `yarn.lock`. Do we need additional `package-lock.json` or was it generated accidentally by using a command like `npm install` instead of `yarn`?

tomaszslabon:
> I see that `package-lock.json` was generated. We already have `yarn.lock`. Do we need additional `package-lock.json` or was it generated accidentally by using a command like `npm install` instead of `yarn`?

Yes, it was generated accidentally. I removed it in d86efc3590a5491b78d717a5887466b08995b5e3.

hardhat.config.js
```diff
@@ -13,8 +14,8 @@ module.exports = {
   networks: {
     hardhat: {
       forking: {
-        // forking is enabled only if FORKING_URL env is provided
-        enabled: !!process.env.FORKING_URL,
+        // forking is enabled only if FORKING_BLOCK is provided (i.e. resetFork was called)
+        enabled: !!process.env.FORKING_BLOCK,
```
dimpar, line 18: What is the reason for changing this to `FORKING_BLOCK`? In all the system tests we set the forking block inside the code. So far it was enough to `FORKING_URL=<url> yarn test:system` without providing `FORKING_BLOCK`.

package.json
```diff
@@ -16,11 +16,13 @@
     "prettier-plugin-solidity": "^1.0.0-beta.2 ",
     "solhint": "^3.3.4",
     "solhint-config-keep": "github:keep-network/solhint-config-keep",
+    "solidity-coverage": "^0.7.16",
     "typescript": "^3.7.0"
   },
   "name": "@keep-network/coverage-pool",
   "scripts": {
     "build": "hardhat compile",
+    "coverage": "NODE_ENV=system-test hardhat coverage",
```
dimpar, line 25: If we manage to fix coverage for system tests, then how about splitting coverage to unit & system, ie
```
    "test:coverage": "yarn test:unit:coverage && yarn test:system:coverage",
    "test:unit:coverage": "hardhat coverage",
    "test:system:coverage": "NODE_ENV=system-test hardhat coverage --testfiles './test/system/*.test.js'",
```

hardhat.config.js
tomaszslabon: We used to run unit tests and system tests separately, so we could rely on `FORKING_URL` to indicate that we want to enable forking or not (we passed `FORKING_URL` for `system tests` and didn't pass it for `unit tests`).

However, to generate coverage reports we need to run **both** unit tests and system tests in one command (some of our contracts are tested in system tests only). So when executing the command, `FORKING_URL` needs to be known (because system tests need it). That's why we cannot rely on whether it is passed or not (it's always passed).

But I've noticed that `FORKING_BLOCK` is only set for system tests (when function `resetFork` is used) and I used it to distinguish between unit tests and system tests.

Also note that if we enabled forking for unit tests, they would run extremely slow (even though `resetFork` is not used in them, the field `enabled` set to `true` is enough). I tried that and they ran 15-20 min or more.

package.json
tomaszslabon: We can do that, but I'm not sure what happens if we run coverage for unit tests and system tests separately (`hardhat coverage` called twice). The generated report will probably contain only coverage from system tests (unit test coverage will be overwritten).

hardhat.config.js
dimpar: I think it boils down to a question if we can fix coverage for system tests. When running coverage for system tests only the result table is completely wrong, shows 100% coverage only for the interfaces and some stubs. There is no notion of coverage of the actual code, even though we test it. After some experiments, it looks like the problem relies in `resetFork()` and `hardhat_reset` method. The problem is known and someone asked very similar question on [gitter](https://gitter.im/sc-forks/solidity-coverage?at=5fdd99ecce40bd3cdbecc5dd) (Pls search for "Quick question, I'm running solidity coverage with hardhat.", cause I can't link exactly his question). The same person suggested to use `evm_snapshot` and `evm_revert` instead.
So if we can fix coverage for system tests, then we can decide whether we should leave or change to what you've proposed. Same for my [comment](https://github.com/keep-network/coverage-pools/pull/128#discussion_r661497775) below.

dimpar (approved):

PR #130, opened 2021-06-29
Add an installation script

r-czajkowski:
Depends on: #103

This PR adds an installation script that:
* installs yarn dependencies,
* deploys coverage pools contracts,
* creates a symbolic link that points to the deployed contract artifacts. This will help to keep consistency between `keep-network` packages. All `keep-network` packages store the contract artifacts in `artifacts` dir eg `@keep-network/keep-core/artifacts/<contract>.json`.

This script will help to run the Keep Token Dashboard locally.

Here we also get rid of the imports that use the relative path to the contract artifact. Instead of this we should use the Hardhat helpers:

**To get the Ethers Contract instance at a given address:**
```
const contract = await ethers.getContractAt(contractName, address, signer)
```

**To get the contract artifact:**
```
const hre = require("hardhat")
const contract = hre.artifacts.readArtifactSync(contractName)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 node_modules/
-artifacts/
+artifacts
+build/
 cache/
 typechain/
 deployments/
--- a/deploy/07_deploy_coverage_pool.js
+++ b/deploy/07_deploy_coverage_pool.js
@@ -9,7 +9,7 @@ module.exports = async ({ getNamedAccounts, deployments }) => {
     args: [AssetPool.address],
     log: true,
   })
-  
+
   const assetPool = await ethers.getContractAt("AssetPool", AssetPool.address)
   const underwriterToken = await ethers.getContractAt(
     "UnderwriterToken",
--- a/deploy/07_deploy_coverage_pool.js
+++ b/deploy/07_deploy_coverage_pool.js
@@ -2,12 +2,22 @@ module.exports = async ({ getNamedAccounts, deployments }) => {
   const { deploy } = deployments
   const { deployer } = await getNamedAccounts()
   const AssetPool = await deployments.get("AssetPool")
+  const UnderwriterToken = await deployments.get("UnderwriterToken")
 
-  await deploy("CoveragePool", {
+  const CoveragePool = await deploy("CoveragePool", {
     from: deployer,
     args: [AssetPool.address],
     log: true,
   })
+  
+  const assetPool = await ethers.getContractAt("AssetPool", AssetPool.address)
+  const underwriterToken = await ethers.getContractAt(
+    "UnderwriterToken",
+    UnderwriterToken.address
+  )
+
+  await assetPool.transferOwnership(CoveragePool.address)
+  await underwriterToken.transferOwnership(AssetPool.address)
 }
 
 module.exports.tags = ["CoveragePool"]
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -1,0 +1,82 @@
+#!/bin/bash
+set -euo pipefail
+
+LOG_START='\n\e[1;36m'  # new line + bold + cyan
+LOG_END='\n\e[0m'       # new line + reset
+DONE_START='\n\e[1;32m' # new line + bold + green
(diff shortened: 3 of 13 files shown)

### CODE REVIEW

File: deploy/07_deploy_coverage_pool.js [hunk 1]
```diff
@@ -2,12 +2,22 @@ module.exports = async ({ getNamedAccounts, deployments }) => {
...
 
-  await deploy("CoveragePool", {
+  const CoveragePool = await deploy("CoveragePool", {
     from: deployer,
     args: [AssetPool.address],
     log: true,
   })
+
+  const assetPool = await ethers.getContractAt("AssetPool", AssetPool.address)
+  const underwriterToken = await ethers.getContractAt(
```
nkuba, line 14: We can simplify it to:
```js
  const assetPool = await ethers.getContract("AssetPool")
  const underwriterToken = await ethers.getContract("UnderwriterToken")
```

But this will also require some configuration. According to [installation instructions](https://github.com/wighawag/hardhat-deploy#installation) we need to:
1. Run `npm install --save-dev  @nomiclabs/hardhat-ethers@npm:hardhat-deploy-ethers ethers`.
2. Add `require("@nomiclabs/hardhat-ethers")` in `hardhat.config.js`.

It's not a big deal but it may be worth doing it now to establish a better pattern for the future scripts enhancement.

scripts/install.sh
```diff
@@ -0,0 +1,82 @@
...
+printf "${LOG_START}Migrating contracts...${LOG_END}"
+KEEP_TOKEN_ADDRESS=$KEEP_TOKEN_ADDRESS \
+TBTC_TOKEN_ADDRESS=$TBTC_TOKEN_ADDRESS \
+TBTC_DEPOSIT_TOKEN_ADDRESS=$TBTC_DEPOSIT_TOKEN_ADDRESS \
+UNISWAP_ROUTER_V2_ADDRESS=$UNISWAP_ROUTER_V2_ADDRESS \
+INITIAL_SWAP_STRATEGY=$INITIAL_SWAP_STRATEGY \
+yarn deploy --reset --network $NETWORK
+
+printf "${LOG_START}Creating links...${LOG_END}"
+ln -sf "deployments/${NETWORK}" artifacts
```
nkuba, line 80: I'm not sure if we should use the same name of the directory as we already use for Truffle-style deployment artifacts. But it's out of the scope of this PR, so let's leave it as is.

nkuba (commented):
Minor comments

.gitignore
```diff
@@ -1,5 +1,6 @@
 node_modules/
-artifacts/
+artifacts
```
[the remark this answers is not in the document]
r-czajkowski: Yes it was.`artifacts` is a [symbolic link](https://github.com/keep-network/coverage-pools/pull/130/files#diff-e076e43cb817ecc8b3c2ed18cd58baa50544ecc66293ed6e61129c43ca6a09ccR80), not a directory so we need to remove `/`.

deploy/07_deploy_coverage_pool.js
r-czajkowski: I've played with ^, but unfortunately, tests fail after installation. I think there is too much magic around dependencies and the `ethers.getSigner` function from the `hardhat-deploy-ethers` is not compatible with the same function from the ` @nomiclabs/hardhat-ethers`.

r-czajkowski: For each invocation `ethers.getSigner` in tests with updated hardhat config I received `Error: need to specify address`.

nkuba: Thanks for trying! The current solution is super fine, if we can't improve it easily I don't want to block this PR. We will handle it in a follow-up.

nkuba (approved):

deploy/07_deploy_coverage_pool.js
nkuba: I managed to get it done in https://github.com/keep-network/coverage-pools/pull/143 :tada:

PR #134, opened 2021-06-29
[System test] Liquidation of undercollateralized deposit

tomaszslabon:
This PR adds an assertion that checks if a deposit in liquidation due to `undercollateralization` can be liquidated vie cov pools.

The deposit in question is [0xfc9c50fd44879bd7085edd311bc8e2b7d3e41595](https://allthekeeps.com/deposit/0xfc9c50fd44879bd7085edd311bc8e2b7d3e41595).

The flow for deposit liquidation was already in the test, it just needed an assertion.


dimpar:
We actually do have undercollateralized deposit in our tests already, which happens to be the same one as in this PR. Pls see [surplus-for-auction.test.js](https://github.com/keep-network/coverage-pools/blob/main/test/system/surplus-for-auction.test.js#L21) and [constants](https://github.com/keep-network/coverage-pools/blob/main/test/system/constants.js#L16). Even though this deposit is liquidated through the "surplus" flow, I think we can maybe just add an assertion that it is in "liquidated" state after the cov pool purchases it and not add a separate system test for that. WDYT?

tomaszslabon:
> We actually do have undercollateralized deposit in our tests already, which happens to be the same one as in this PR. Pls see [surplus-for-auction.test.js](https://github.com/keep-network/coverage-pools/blob/main/test/system/surplus-for-auction.test.js#L21) and [constants](https://github.com/keep-network/coverage-pools/blob/main/test/system/constants.js#L16). Even though this deposit is liquidated through the "surplus" flow, I think we can maybe just add an assertion that it is in "liquidated" state after the cov pool purchases it and not add a separate system test for that. WDYT?

I agree.
@pdyrag, WDYT?

tomaszslabon:
> We actually do have undercollateralized deposit in our tests already, which happens to be the same one as in this PR. Pls see [surplus-for-auction.test.js](https://github.com/keep-network/coverage-pools/blob/main/test/system/surplus-for-auction.test.js#L21) and [constants](https://github.com/keep-network/coverage-pools/blob/main/test/system/constants.js#L16). Even though this deposit is liquidated through the "surplus" flow, I think we can maybe just add an assertion that it is in "liquidated" state after the cov pool purchases it and not add a separate system test for that. WDYT?

@dimpar 
I agree, done in 603a1c57159959bc29cc030c551fd0034adac096.

dimpar (approved):

Issue #126, opened 2021-06-27
Recommend Auction length based on Cov Pool TVL


pdyraga:
Thoroughly document auction length parameter meaning and importance. Explore scenarios, propose various auction length values based on asset pool TVL.

We have already done some work [here](https://docs.google.com/spreadsheets/d/1vsMZcvRAEX180pxnxk0iC9xaL4hEw6Lx-96UI_1ZJ3A/edit?usp=sharing) but this is not something that can be passed to Community Multisig in the current state. To consider this issue as Done, the documentation needs to be comprehensive, reviewed by Community Multisig members and adjusted based on their feedback.

dimpar:
> For 100% case:
> 
> * 0.01 TBTC lot size becomes profitable after ~73 seconds

For 100% case:
* 0.01 TBTC lot size becomes profitable after ~18 seconds

dimpar:
> How one-week auction length looks for the case when BTC/ETH stays the same but 1 ETH = 2500 KEEP for 50% and 100% of KEEP migrated between KEEP-only pool and coverage pool.

For 50% case:
- 0.01 TBTC lot size becomes profitable after ~12 seconds
- 0.1 TBTC lot size becomes profitable after ~2 minutes
- 0.2 TBTC lot size becomes profitable after ~4 minutes
- 0.5 TBTC lot size becomes profitable after ~10 minutes
- 1 TBTC lot size becomes profitable after ~20 minutes
- 5 TBTC lot size becomes profitable after ~97 minutes
- 10 TBTC lot size becomes profitable after ~193 minutes

For 100% case:
- 0.01 TBTC lot size becomes profitable after ~6 seconds
- 0.1 TBTC lot size becomes profitable after ~1 minutes
- 0.2 TBTC lot size becomes profitable after ~2 minutes
- 0.5 TBTC lot size becomes profitable after ~5 minutes
- 1 TBTC lot size becomes profitable after ~10 minutes
- 5 TBTC lot size becomes profitable after ~49 minutes
- 10 TBTC lot size becomes profitable after ~97 minutes

pdyraga:
Based on this information I think a one-week auction length is reasonable for the beginning. What do you think @dimpar? If we agree here, I'll update the deployment scenario and once we confirm with the community multisig they understand the spreadsheet and how the auction length works, we can close this issue.

dimpar:
[...]

PR #139, opened 2021-07-01
Updating docs around auction length parameter

dimpar:
Refs https://github.com/keep-network/coverage-pools/issues/126

This PR describes "auction length" parameter and how it affects portion on offer over time of an Asset Pool at the given moment. [Spreadsheet](https://docs.google.com/spreadsheets/d/1ukn45bmPFL8EAl87rYz_NQgBX_n2hxrDD0oyag33ZCY/edit#gid=412009494) simulates different scenarios of portions of offer based on factors like coverage pool TVL and auction length.

The goal is to find a "sweet spot" for auction length, not making it too short (which leads to big sums of asset pools become available in a short time) and not making it too long (which leads to bidders waiting for too long until it will makes sense for them to bid on an auction). One of the collateral tokens in asset pools will be KEEP token. Assumption is that the cov pool TVL for KEEP collateral will be between `50-100%` of staked tokens in [KEEP-only pool](https://etherscan.io/address/0xdF00daC2Be1250cF62cBFc617EE7bC45C0016c87#readContract). Currently the staked KEEP tokens in KEEP-only pool are valued for around `~6k` in ETH. It seems that if we pick `auction length` param as `168h` which is `7days`, then for the lowest deposit lot size `0.01TBTC` it will be profitable to bid after `~18sec` and for the highest lot size `10TBTC` it will be profitable to bid after `~4h30min`. `168h` also gives us a safe margin if TVL goes `+-50%`


docs/design.adoc
```diff
@@ -290,6 +290,16 @@ being liquidated outside of the coverage pool.
...
+
+** first step is calling `beginAuctionLengthUpdate(newAuctionLength)`, which sets a new auction length parameter.
+** second call can be done after the governance delay has passed, which is currently set to 12h. After this time, one should call `finalizeAuctionLengthUpdate()` to complete auction length parameter update.
+
+The longer the auction length is, the less portion (collateral tokens) of an asset pool is on offer at the given moment from auction start time.
+
```
tomaszslabon, line 302: Could you please format it, so that one line is at max 80 chars long?

docs/design.adoc
```diff
@@ -290,6 +290,16 @@ being liquidated outside of the coverage pool.
...
 the fixed amount reward takes precedence.
 
+=== Auction length upgradability
+
+The default value of "Auction Length" is set during deployment of Risk Manager 
+contract. This parameter is one of the factors that determines the value of asset pool portion being on offer at the given time. It should be updated carefully, because selecting too short auction length value when the coverage pool TVL is big can lead to significant portions being on offer in a very short period of time. It means draining the pool quickly and making auction bidders very profitable. On the other hand auction length should not be very long, because bidders will have to wait too much time before it will make sense for them to take an offer on an auction. Please refer to auction length simulation https://docs.google.com/spreadsheets/d/1ukn45bmPFL8EAl87rYz_NQgBX_n2hxrDD0oyag33ZCY/edit?usp=sharing[spreadsheet] based on different pool's TVL. The detailed description of the auction length parameter can be found https://github.com/keep-network/coverage-pools/blob/main/contracts/RiskManagerV1.sol#L104[here]. Updating mechanism is done in two steps:
```
tomaszslabon, line 296: factors that determines -> factors that determine

docs/design.adoc
dimpar: Sure, VCS wrapped it for me while typing and I didn't notice it. Done in 31dfa90

docs/design.adoc
dimpar: Done in 31dfa90

docs/design.adoc
```diff
@@ -290,6 +290,31 @@ being liquidated outside of the coverage pool.
 If both fixed amount and percentage values are set and bigger than zero,
 the fixed amount reward takes precedence.
 
+=== Auction length upgradability
```
pdyraga, line 293: Can we name this section "Auction length value" so that it is not confused with the upgradeability mechanism between asset pools?

docs/design.adoc
```diff
@@ -290,6 +290,31 @@ being liquidated outside of the coverage pool.
...
+
+The default value of "Auction Length" is set during deployment of Risk Manager
```
pdyraga, line 295: s/"Auction Length"/auction length IMO, it is just a value, it is not a component or role name.

pdyraga (commented):
Looks good to me, just two comments.

I do not think this PR closes the issue, we need to do some more work; from the issue description:

> To consider this issue as Done, the documentation needs to be comprehensive, reviewed by Community Multisig members and adjusted based on their feedback.

docs/design.adoc
dimpar: Sure, done in 1663b7f

pdyraga (approved):

PR #142, opened 2021-07-02
CI workflow to test deployment scripts

nkuba:
This workflow will run deployment scripts against the local hardhat network instance. This is to test the scripts.

Depends on https://github.com/keep-network/coverage-pools/pull/103

### DIFF
--- a/.github/workflows/contracts.yml
+++ b/.github/workflows/contracts.yml
@@ -16,6 +16,7 @@ jobs:
       - uses: actions/setup-node@v2
         with:
           node-version: "12.x"
+          cache: "yarn"
 
       - name: Cache yarn directory
         uses: c-hive/gha-yarn-cache@v1
@@ -42,6 +43,7 @@ jobs:
       - uses: actions/setup-node@v2
         with:
           node-version: "12.x"
+          cache: "yarn"
 
       - name: Cache yarn directory
         uses: c-hive/gha-yarn-cache@v1
@@ -60,6 +62,7 @@ jobs:
       - uses: actions/setup-node@v2
         with:
           node-version: "12.x"
+          cache: "yarn"
 
       - name: Cache yarn directory
         uses: c-hive/gha-yarn-cache@v1
@@ -78,6 +81,7 @@ jobs:
       - uses: actions/setup-node@v2
         with:
           node-version: "12.x"
+          cache: "yarn"
 
       - uses: actions/setup-python@v2
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/contracts.yml
```diff
@@ -34,6 +34,24 @@ jobs:
           FORKING_URL: ${{ secrets.MAINNET_ETH_HOSTNAME }}
         run: yarn test:system
 
+  contracts-deployment-test:
```
michalinacienciala, line 37:
```suggestion
  contracts-deployment:
```
Could we drop the `-tests` from the name? I don't see any tests performed in this job.

nkuba: It is a test run of the deployment scripts on local hardhat network instance. I know the name is confusing. Maybe I'll update it to `contracts-deployment-dry-run`?

michalinacienciala: Oh, I see. Im fine with your proposition of the change.

.github/workflows/contracts.yml
```diff
@@ -34,6 +34,24 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: actions/setup-node@v2
+        with:
+          node-version: "12.x"
+
+      - name: Cache yarn directory
+        uses: c-hive/gha-yarn-cache@v1
```
michalinacienciala, line 47: We could try using this functionality in `setup-node` action for caching:
```suggestion
      - uses: actions/setup-node@v2
        with:
          node-version: "12.x"
          cache: "yarn"
```
(Also applies to other jobs in the workflow)

michalinacienciala (approved):
Looks good to me, but I included one optional suggestion

.github/workflows/contracts.yml
```diff
@@ -34,6 +35,25 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: actions/setup-node@v2
+        with:
+          node-version: "12.x"
+          cache: "yarn"
+
+      - name: Cache yarn directory
+        uses: c-hive/gha-yarn-cache@v1
```
michalinacienciala, line 49: I think this duplicates the
```
      - uses: actions/setup-node@v2
        with:
          node-version: "12.x"
          cache: "yarn"
```
We can get rid of it (here and in other places in the workflow).

michalinacienciala (changes requested):

michalinacienciala (approved):

PR #143, opened 2021-07-02
Use hardhat-deploy-ethers plugin

nkuba:
The plugin `hardhat-deploy-ethers` from creators of `hardhat-deploy` plugin we use for deployment. The plugin enhances original [@nomiclabs/hardhat-ethers](https://www.npmjs.com/package/@nomiclabs/hardhat-ethers) with additional functionalities that will simplify our code, e.g. `await ethers.getContract("ContractName")` function.

See https://github.com/wighawag/hardhat-deploy-ethers for details.

According to API (https://github.com/wighawag/hardhat-deploy-ethers) `getSigner` expects an address as parameter. To get a signer by index we switched to `getSigners()` function.

Depends on: https://github.com/keep-network/coverage-pools/pull/103

### DIFF
--- a/hardhat.config.js
+++ b/hardhat.config.js
@@ -1,4 +1,5 @@
 require("@nomiclabs/hardhat-waffle")
+require("@nomiclabs/hardhat-ethers")
 require("hardhat-gas-reporter")
 require("hardhat-deploy")
 require("solidity-coverage")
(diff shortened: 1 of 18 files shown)


nkuba:
As we updated our deployment scripts and no longer use `await ethers.getContract("ContractName")` function we don't need this PR. I'm closing it now. We can always get back to it in the future.

Issue #133, opened 2021-06-29
Division rounding may affect minting of underwriter tokens


ggrieco-tob:
```
Severity: Medium						Difficulty: High
Type: Arithmetic						
Target: AssetPool.sol
```

Users may not receive underwriter tokens in exchange for their deposits if the total supply amount is manipulated by another user through a large "donation."

The `_deposit` function calculates the amount of tokens to be minted:

https://github.com/keep-network/coverage-pools/blob/main/contracts/AssetPool.sol#L464-L481

Essentially, the amount of tokens that a user will receive depends on the total supply and the number of collateral tokens holded by the contract. 

However, if a user makes a large “donation” to the contract address, increasing the total supply, other users may not receive tokens after making deposits.

**Exploit Scenario**
1. Alice deploys the underwriter token. She makes a minimum initial deposit to obtain one unit of tokens, so the `totalSupply` is one. 

2. Eve wants to block Alice's token, so she transfers a large number of tokens to the contract address, increasing the `collateralBalance` value without increasing the `totalSupply`.

3. A user calls deposit in order to obtain tokens. Because their deposit is smaller than Eve’s, they do not receive any tokens, even though the transaction succeeds. As a result, the contract has an increased amount of collateral tokens, but the same number of tokens minted (one).

If more users deposit small amounts, they will not receive any token in return. The problem will get worse until a user with enough money makes a large deposit.

**Recommendations**
Short term, considering adding `require(toMint > 0)` to mitigate this issue and document the token initialization process to ensure that the deployer’s first deposit is large enough to prevent corner cases in which users do not receive tokens after making deposits.

Long term, review the list of global invariants to make sure there is no way to break them.

tomaszslabon:
@pdyraga
I added a [fix](https://github.com/keep-network/coverage-pools/pull/145).
Once it's merged, we will be able to close this issue.

PR #145, opened 2021-07-05
Check amount of underwriter tokens that are minted

tomaszslabon:
This PR adds a check for the amount of minted underwriter tokens when collateral tokens are deposited so that the `deposit` transaction reverts if the amount of COV to mint is zero.

The check can fail in the following scenario:
1) deposit a small amount of collateral tokens (so that the supply of underwriter tokens is at least `1`)
2) send (not deposit) a bigger amount of collateral tokens to the asset pool
3) deposit less collateral tokens that sent to the contract in the previous step

Closes https://github.com/keep-network/coverage-pools/issues/133.

### DIFF
--- a/contracts/AssetPool.sol
+++ b/contracts/AssetPool.sol
@@ -363,7 +363,7 @@ contract AssetPool is Ownable, IAssetPool {
     }
 
     /// @notice Lets the contract owner to begin an update of withdrawal timeout
-    ///         parmeter value. The withdrawal timeout is the time the
+    ///         parameter value. The withdrawal timeout is the time the
     ///         underwriter has - after the withdrawal delay passed - to
     ///         complete the withdrawal. The change needs to be finalized with
     ///         a call to finalizeWithdrawalTimeoutUpdate after the required
--- a/contracts/AssetPool.sol
+++ b/contracts/AssetPool.sol
@@ -478,7 +478,7 @@ contract AssetPool is Ownable, IAssetPool {
         } else {
             toMint = (amount * covSupply) / collateralBalance;
         }
-        require(toMint > 0, "Minted amount would be 0");
+        require(toMint > 0, "Minted tokens amount must be greater than 0");
 
         underwriterToken.mint(depositor, toMint);
         collateralToken.safeTransferFrom(depositor, address(this), amount);
--- a/contracts/AssetPool.sol
+++ b/contracts/AssetPool.sol
@@ -478,6 +478,8 @@ contract AssetPool is Ownable, IAssetPool {
         } else {
             toMint = (amount * covSupply) / collateralBalance;
         }
+        require(toMint > 0, "Minted amount would be 0");
+
         underwriterToken.mint(depositor, toMint);
         collateralToken.safeTransferFrom(depositor, address(this), amount);
     }
--- a/test/AssetPool.test.js
+++ b/test/AssetPool.test.js
@@ -228,28 +228,26 @@ describe("AssetPool", () => {
       })
     })
 
-    context(
-      "when collateral tokens sent to asset pool outside of cov pools",
-      () => {
-        beforeEach(async () => {
-          // Deposit some small amount of collateral tokens so that the amount
-          // of underwriter tokens is not 0
-          await assetPool.connect(underwriter1).deposit(1)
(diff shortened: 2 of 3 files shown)


contracts/AssetPool.sol
```diff
@@ -478,6 +478,8 @@ contract AssetPool is Ownable, IAssetPool {
         } else {
             toMint = (amount * covSupply) / collateralBalance;
         }
+        require(toMint > 0, "Minted amount would be 0");
```
dimpar, line 481: How about "Minted tokens amount must be greater than 0"?

dimpar: Alternative solution but probably an overkill is to keep track of all collateral tokens through `deposit()` and `withdrawal()` functions. Even if someone transfer `collateralToken` directly, it won't affect the calculation around `toMint` tokens.

tomaszslabon: Fixed in 1112ce07ad93245fc41dd9c69431d2b9c6e9a1fe.

tomaszslabon: As for alternative solution (keeping track of all collateral tokens in `asset pool` via `deposited` and `withdrawal`:
I thought about, but came to the conclusion that it would indeed be an overkill.

test/AssetPool.test.js
```diff
@@ -227,6 +227,29 @@ describe("AssetPool", () => {
         ).to.be.closeTo("45454545454545454545", assertionPrecision)
       })
     })
+
+    context(
+      "when collateral tokens sent to asset pool outside of cov pools",
```
pdyraga, line 232: I am not sure if the name of this test explains well what we are trying to check. It is fine if someone does a plain transfer, it's how ETH->KEEP transition from purchased bonds will work.

tomaszslabon: @pdyraga 
I think it mainly affects how many `underwriter tokens` will be minted for the `underwriter` in exchange for the `collateral tokens` they are depositing. 
Any suggestion how I could name this test?
Maybe `when collateral tokens in asset pool are increased via plain transfer`?

pdyraga: Correct. What I mean is that:
- "when collateral tokens sent to asset pool outside of cov pools"
- "when collateral tokens in asset pool are increased via plain transfer"

are both perfectly valid cases and if we say that the deposit should always revert in such situations, this is wrong because it is not what is actually expected. We do not expect `deposit` to always revert if ETH->KEEP swap happened from purchased signer bonds and that KEEP funded `AssetPool`. Similarly, we do not expect `deposit` to always revert if `RewardsPool` released KEEP tokens and funded `AssetPool`.

We do however expect `deposit` to revert if the amount of COV minted would be 0.

So I suggest we name this scenario:
```
  context("when minting 0 underwriter tokens", () => {
    // (...)
    it("should revert", async () => {
}
```

because this is what we are actually preventing from and asserting.

test/AssetPool.test.js
```diff
@@ -227,6 +227,29 @@ describe("AssetPool", () => {
...
+      () => {
+        beforeEach(async () => {
+          // Deposit some small amount of collateral tokens so that the amount
+          // of underwriter tokens is not 0
+          await assetPool.connect(underwriter1).deposit(1)
+          // Send collateral tokens to the asset pool outside cov pools
```
pdyraga, line 238: Maybe:
```
// Send collateral tokens to the asset pool directly, without minting underwriter tokens
```

The second part is essential to understand why `toMint` is `0`.

pdyraga:
You can add `Closes #133` to the PR description @tomaszslabon and the issue will get closed on this PR gets merged.

test/AssetPool.test.js
tomaszslabon: OK, now I get it.
Fixed in https://github.com/keep-network/coverage-pools/pull/145/commits/61952dd033344000712d1a74a06244b2f38331dc.

test/AssetPool.test.js
tomaszslabon: Fixed in https://github.com/keep-network/coverage-pools/pull/145/commits/61952dd033344000712d1a74a06244b2f38331dc.

pdyraga (approved):

Issue #136, opened 2021-06-30
Constant instead of enum is error-prone


montyly:
Severity: Informational

## Description
The codebase uses constants for enum-based values. This schema is error-prone and might lead the constant to not be up to date with the enums

For example:
https://github.com/keep-network/coverage-pools/blob/09aa12565ab46a954ab2cb969209e6721bd6b631/contracts/RiskManagerV1.sol#L88-L90

https://github.com/keep-network/tbtc/blob/f8d76ad08a06f1989c9328d81428c4b030226561/solidity/contracts/deposit/DepositStates.sol#L22-L25

## Recommendations
Short term, use an enum in `RiskManagerV1`, and evaluate similar usage of constant in the codebase.

Long term, create interfaces and libraries such that they can be used with multiple codebases and compiler versions

tomaszslabon:
@pdyraga 
I added a [fix](https://github.com/keep-network/coverage-pools/pull/147) in which I replaced constants with enums.
It increased gas consumption a bit and it's slightly more code.

Should we merge it or should we just leave the constants?

pdyraga:
I suggest we leave it as-is. I do agree constants can be error-prone but in this specific case we have tBTC v1 deployed on mainnet, compiled with a specific version of Solidity, and values for those constants can not change. Furthermore, assuming we copy-paste enums, there would be a marginal possibility of Solidity compiler changing how constants are represented and tBTC v1 enums compiled with Solidity 0.5 could not match the ones compiled for coverage pool constants.

PR #147, opened 2021-07-06
Used enums instead of consts for deposit states

tomaszslabon:
This PR replaces constant values with enums to mark deposit states.
It addresses issue https://github.com/keep-network/coverage-pools/issues/136.

It increased (average) gas consumption a little in system tests:
- `takeOffer`: 197936 -> 198034
- `notifyLiquidated`: 82987 -> 83090
- `notifyLiquidation`: 374157 -> 374337
- `swapSignerBondsOnUniswapV2`: 170472 -> 170516

Some other function changed (increased or decreased) gas consumption, but these were minimal changes. 
It wasn't necessary to increase gas consumption expectations in tests.

### DIFF
--- a/contracts/RiskManagerV1.sol
+++ b/contracts/RiskManagerV1.sol
@@ -80,14 +80,16 @@ contract RiskManagerV1 is IRiskManagerV1, Auctioneer, Ownable {
     using SafeERC20 for IERC20;
     using RiskManagerV1Rewards for RiskManagerV1Rewards.Storage;
 
+    enum DepositState {
+        FRAUD_LIQUIDATION_IN_PROGRESS,
+        LIQUIDATION_IN_PROGRESS,
+        LIQUIDATED,
+        OTHER
+    }
+
     /// @notice Governance delay that needs to pass before any risk manager
     ///         parameter change initiated by the governance takes effect.
     uint256 public constant GOVERNANCE_DELAY = 12 hours;
-
-    // See https://github.com/keep-network/tbtc/blob/v1.1.0/solidity/contracts/deposit/DepositStates.sol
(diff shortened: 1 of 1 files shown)


contracts/RiskManagerV1.sol
```diff
@@ -807,10 +809,25 @@ contract RiskManagerV1 is IRiskManagerV1, Auctioneer, Ownable {
...
+        return (state == DepositState.LIQUIDATION_IN_PROGRESS ||
+            state == DepositState.FRAUD_LIQUIDATION_IN_PROGRESS);
+    }
 
-        return (state == DEPOSIT_LIQUIDATION_IN_PROGRESS_STATE ||
-            state == DEPOSIT_FRAUD_LIQUIDATION_IN_PROGRESS_STATE);
+    /// @dev To learn more about deposit states see
+    /// https://github.com/keep-network/tbtc/blob/v1.1.0/solidity/contracts/deposit/DepositStates.sol
+    function uintToState(uint256 state) internal pure returns (DepositState) {
+        if (state == 9) {
```
dimpar, line 821: It feels to me that we still have the same issue with hardcoding states with these `if` statements. We _probably_ can work on adding `DepositStates.sol` contract and calling to check a deposit state but, `tbtc` doesn't go anywhere and these states won't change. Idk, maybe we can leave as it is currently implemented?

tomaszslabon: I have the same feeling, but wanted to check how it would look if we replaced consts with enums.
@pdyraga WDYT?

pdyraga: Yeah, I do agree with @dimpar, will post some more thoughts in the issue.

Issue #135, opened 2021-06-30
Uniswap price manipulation allows profiting from risk manager swaps


montyly:
Severity: High
Difficulty: Low

## Description
The lack of swap price guarantees allow anyone to perform unfair swaps from the risk manager and profit from the arbitrage opportunities created.

Anyone can trigger a swap through uniswap with the risk manager funds:
https://github.com/keep-network/coverage-pools/blob/09aa12565ab46a954ab2cb969209e6721bd6b631/contracts/SignerBondsUniswapV2.sol#L193-L239

There is no proper mechanism to ensure that the swap will be done at a fair market price. If an attacker shifts the spot price on the uniswap pair, he will make the the risk manager perform an unfair swap. The attacker can then benefit from the arbitrage opportunities created to make a profit.

Note that the `isAllowedPriceImpact` is supposed to prevent slippages, but does not protect against an unbalanced pool.

## Exploit Scenario
The risk manager has 10 ether. Eve uses a flash loan to shift the spot price on the uniswap pool. The pool has a large amount of Ether, and few collaterals. `getAmountsOut(10 ether, ..)` returns 0. Eve calls `swapSignerBondsOnUniswapV2` to trigger the swap. She then sells back all the collaterals bought, and profit from the arbitrage opportunity created to make a profit

## Recommendations
Short term, consider to:
- Restrict the call to `swapSignerBondsOnUniswapV2` to trusted users
- Use a price Oracle to ensure the swap is done at a fair price
 
Long term, design your contracts such that you avoid arbitrary users to perform assets swaps on behalf of contracts.

tomaszslabon:
@pdyraga I added a [fix](https://github.com/keep-network/coverage-pools/pull/148) suggested by you.
I think we can close this issue, after we merge the fix.

PR #148, opened 2021-07-07
Approving and unapproving signer bonds swappers

tomaszslabon:
This PR adds functionalities for `approving` and `unapproving` operators that can swap signer bonds on `Uniswap V2`.
Closes https://github.com/keep-network/coverage-pools/issues/135.

### DIFF
--- a/contracts/SignerBondsUniswapV2.sol
+++ b/contracts/SignerBondsUniswapV2.sol
@@ -77,6 +77,9 @@ contract SignerBondsUniswapV2 is ISignerBondsSwapStrategy, Ownable {
     address public immutable assetPool;
     address public immutable collateralToken;
 
+    // Currently approved operators that can swap signer bonds on Uniswap V2
+    mapping(address => bool) public approvedOperators;
+
     // Determines the maximum allowed price impact for the swap transaction.
     // If transaction's price impact is higher, transaction will be reverted.
     // Default value is 100 basis points (1%).
@@ -93,7 +96,22 @@ contract SignerBondsUniswapV2 is ISignerBondsSwapStrategy, Ownable {
     // If false, open auctions are not taken into account.
     bool public revertIfAuctionOpen = true;
 
+    event SignerBondsSwapOperatorApproved(address operator, uint256 timestamp);
+    event SignerBondsSwapOperatorUnapproved(
+        address operator,
+        uint256 timestamp
+    );
     event UniswapV2SwapExecuted(uint256[] amounts);
+
+    /// @notice Reverts if called by a signer bonds swap operator that is not
+    ///         approved
+    modifier onlyApprovedOperator() {
+        require(
+            approvedOperators[msg.sender],
+            "Signer bonds swap operator not approved"
+        );
+        _;
+    }
 
     constructor(IUniswapV2Router _uniswapRouter, CoveragePool _coveragePool) {
         uniswapRouter = _uniswapRouter;
@@ -194,7 +212,7 @@ contract SignerBondsUniswapV2 is ISignerBondsSwapStrategy, Ownable {
     function swapSignerBondsOnUniswapV2(
         IRiskManagerV1 riskManager,
         uint256 amount
-    ) external {
+    ) external onlyApprovedOperator {
         require(amount > 0, "Amount must be greater than 0");
         require(
             amount <= address(riskManager).balance,
@@ -239,6 +257,30 @@ contract SignerBondsUniswapV2 is ISignerBondsSwapStrategy, Ownable {
             );
 
         emit UniswapV2SwapExecuted(amounts);
+    }
+
+    /// @notice Approves the signer bonds swap operator. The change takes effect
+    ///         immediately.
+    /// @dev Can be called only by the contract owner.
+    /// @param operator Operator that will be approved
+    function approveOperator(address operator) external onlyOwner {
+        /* solhint-disable-next-line not-rely-on-time */
+        emit SignerBondsSwapOperatorApproved(operator, block.timestamp);
+        approvedOperators[operator] = true;
+    }
+
+    /// @notice Unapproves the signer bonds swap Operator. The change takes
+    ///         effect immediately.
+    /// @dev Can be called only by the contract owner.
+    /// @param operator Operator that will be unapproved
+    function unapproveOperator(address operator) external onlyOwner {
+        require(
+            approvedOperators[operator],
+            "Signer swap operator is not approved"
+        );
+        /* solhint-disable-next-line not-rely-on-time */
+        emit SignerBondsSwapOperatorUnapproved(operator, block.timestamp);
+        delete approvedOperators[operator];
     }
 
     /// @notice Checks the price impact of buying a given amount of tokens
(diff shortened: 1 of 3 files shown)


test/system/uniswap-v2.test.js
```diff
@@ -60,14 +60,35 @@ describeFn("System -- swap signer bonds on UniswapV2", () => {
         expect(balance).to.equal(ethers.utils.parseEther("20"))
       })
     })
+
+    describe("signer bonds uniswap v2", () => {
```
dimpar, line 64: It looks like this test belongs to unit tests, because we're checking just a small functionality of approved "operators". Do you think we can create a unit test and move it there?

pdyraga: Agreed, this should live in unit tests.

contracts/SignerBondsUniswapV2.sol
```diff
@@ -93,8 +96,23 @@ contract SignerBondsUniswapV2 is ISignerBondsSwapStrategy, Ownable {
     // If false, open auctions are not taken into account.
     bool public revertIfAuctionOpen = true;
 
+    event SignerBondsSwapOperatorApproved(address operator, uint256 timestamp);
```
pdyraga, line 99: Do we need to add `timestamp` parameter to these events? The timestamp can be read from block add we add `timestamp` only to `Begin*` governance events just because the timestamp is the part of the business logic and not a metadata - that is, when the governance delay started.

contracts/SignerBondsUniswapV2.sol
```diff
@@ -77,6 +77,9 @@ contract SignerBondsUniswapV2 is ISignerBondsSwapStrategy, Ownable {
     address public immutable assetPool;
     address public immutable collateralToken;
 
+    // Currently approved operators that can swap signer bonds on Uniswap V2
```
pdyraga, line 80: Can we drop "Currently"? We do not have a list of pending approvals, so I think *currently* is misleading.

contracts/SignerBondsUniswapV2.sol
```diff
@@ -77,6 +77,9 @@ contract SignerBondsUniswapV2 is ISignerBondsSwapStrategy, Ownable {
...
+    mapping(address => bool) public approvedOperators;
```
pdyraga, line 81: I would consider naming them differently than operators given that "operator" is a very specific role in Keep Network and contracts. Maybe `approvedSwappers`?

contracts/SignerBondsUniswapV2.sol
tomaszslabon: I removed the timestamps in https://github.com/keep-network/coverage-pools/pull/148/commits/cee62f0dba2afa4442e29eff73ffec2170ac3dd2.

contracts/SignerBondsUniswapV2.sol
tomaszslabon: I removed the comment entirely.

contracts/SignerBondsUniswapV2.sol
tomaszslabon: I changed the name of mapping to `approvedSwappers` in https://github.com/keep-network/coverage-pools/pull/148/commits/cee62f0dba2afa4442e29eff73ffec2170ac3dd2

test/system/uniswap-v2.test.js
tomaszslabon: @pdyraga @dimpar 
I moved the tests to unit tests [`SignerBondsUniswapV2.test.js`](https://github.com/keep-network/coverage-pools/blob/only-approved-signer-swap-operators/test/SignerBondsUniswapV2.test.js).
However, I skipped one testcase that was affected by this PR (the happy path for `swapSignerBondsOnUniswapV2`), as it's tested in a system test (it was hard to test it in unit tests, that's why we decided to write system tests for it). 

There are more simple functions in `SignerBondsUniswapV2` that could be tested in unit tests, but are unrelated to this PR (e.g. `setMaxAllowedPriceImpact`, `setSlippageTolerance`). Should I add them (in a separate PR)?

test/SignerBondsUniswapV2.test.js
```diff
@@ -0,0 +1,166 @@
...
+          .connect(governance)
+          .approveSwapper(swapper.address)
+      })
+
+      it("should approve the swapper", async () => {
+        expect(await signerBondsUniswapV2.approvedSwappers(swapper.address)).to
+          .be.true
+      })
+
+      it("should emit RiskManagerUnapproved event", async () => {
```
pdyraga, line 101: s/RiskManagerUnapproved/SignerBondsSwapperApproved

contracts/SignerBondsUniswapV2.sol
```diff
@@ -241,6 +254,25 @@ contract SignerBondsUniswapV2 is ISignerBondsSwapStrategy, Ownable {
...
+        emit SignerBondsSwapperApproved(swapper);
+        approvedSwappers[swapper] = true;
+    }
+
+    /// @notice Unapproves the signer bonds swapper. The change takes effect
+    ///         immediately.
+    /// @dev Can be called only by the contract owner.
+    /// @param swapper Swapper that will be unapproved
+    function unapproveSwapper(address swapper) external onlyOwner {
+        require(approvedSwappers[swapper], "Signer swapper is not approved");
```
pdyraga, line 271: Why *signer swapper*? Should probably be *signer bonds swapper*

test/SignerBondsUniswapV2.test.js
```diff
@@ -0,0 +1,166 @@
...
+          .connect(governance)
+          .unapproveSwapper(swapper.address)
+      })
+
+      it("should unapprove the swapper", async () => {
+        expect(await signerBondsUniswapV2.approvedSwappers(swapper.address)).to
+          .be.false
+      })
+
+      it("should emit RiskManagerUnapproved event", async () => {
```
pdyraga, line 147: s/RiskManagerUnapproved/SignerBondsSwapperUnapproved

test/SignerBondsUniswapV2.test.js
```diff
@@ -0,0 +1,166 @@
...
+      })
+    })
+  })
+
+  describe("swapSignerBondsOnUniswapV2", () => {
+    context("when swapper is not approved", () => {
+      it("should revert", async () => {
+        await expect(
+          signerBondsUniswapV2
+            .connect(swapper)
```
pdyraga, line 160: Nitpick but we could use `thirdParty` here to be even more explicit.

tomaszslabon: Done in https://github.com/keep-network/coverage-pools/pull/148/commits/8b06db86517b95422ace7846320b14a1cbaee844.

test/SignerBondsUniswapV2.test.js
tomaszslabon: Fixed in https://github.com/keep-network/coverage-pools/pull/148/commits/8b06db86517b95422ace7846320b14a1cbaee844

contracts/SignerBondsUniswapV2.sol
tomaszslabon: Fixed in https://github.com/keep-network/coverage-pools/pull/148/commits/8b06db86517b95422ace7846320b14a1cbaee844

test/SignerBondsUniswapV2.test.js
tomaszslabon: Copy-paste error, fixed in https://github.com/keep-network/coverage-pools/pull/148/commits/8b06db86517b95422ace7846320b14a1cbaee844

pdyraga (approved):

PR #149, opened 2021-07-07
Coverage Pool Beneficiary for coverage pool rewards contract

dimpar:
Adding `CoveragePoolBeneficiary` contract to fund coverage rewards pool with rewards tokens. For the funding purposes the existing [BatchedPhasedEscrow](https://etherscan.io/address/0x1b8e50ec9fbf844c3671bc178df8eadfcff831ca) will be used to set up beneficiary, ie.`CoveragePoolBeneficiary`. The flow should be the following:
`BatchedPhasedEscrow` -> `CoveragePoolBeneficiary` -> `RewardsPool`


contracts/PhasedEscrow.sol
```diff
@@ -0,0 +1,82 @@
...
+        rewardsPool.topUpReward(amount);
+    }
+
+    /// @notice Funds the escrow by transferring all of the approved tokens
+    ///         to the escrow.
+    function receiveApproval(
+        address _from,
+        uint256 _value,
+        address _token,
+        bytes memory
```
tomaszslabon, line 77: `bytes` is unused here. Is this a standard to include it in `reveiveApproval` functions or can we remove it?

tomaszslabon (commented):
Will it be used in unit/system tests to check if it works?

contracts/CoveragePoolBeneficiary.sol
```diff
@@ -0,0 +1,42 @@
...
+
+interface IBeneficiaryContract {
+    function __escrowSentTokens(uint256 amount) external;
+}
+
+contract CoveragePoolBeneficiary is Ownable, IBeneficiaryContract {
+    using SafeERC20 for IERC20;
+
+    IERC20 public immutable token;
+    IRewardsPool public immutable rewards;
```
pdyraga, line 31: I would interact with `RewardsPool` directly. External contracts do not need to talk to `RewardsPool` and this beneficiary is specific to a single `RewardsPool` implementation.

contracts/CoveragePoolBeneficiary.sol
```diff
@@ -0,0 +1,42 @@
...
+
+pragma solidity 0.8.4;
+
+import "./interfaces/IRewardsPool.sol";
+
+import "@openzeppelin/contracts/access/Ownable.sol";
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+
+interface IBeneficiaryContract {
```
pdyraga, line 23: Let's reference https://github.com/keep-network/keep-core/blob/main/solidity/contracts/PhasedEscrow.sol in a comment.

pdyraga (commented):
Tagging @nkuba to check if there is anything that could complicate the deployment. We will use an existing `BatchedPhasedEscrow` deployed at `0x1b8e50ec9fbf844c3671bc178df8eadfcff831ca` so it should be just a matter of deploying `CoveragePoolBeneficiary` and approving this beneficiary by the appropriate multisig outside of the deployment script.

contracts/PhasedEscrow.sol
pdyraga: Yes, this interface is an industry standard though I think it was never included in an EIP that got finalized.

contracts/CoveragePoolBeneficiary.sol
dimpar: The reason why I added an interface is because of [RewardsPoolV2](https://github.com/keep-network/coverage-pools/blob/main/contracts/v2/RewardsPoolV2.sol#L81) contract. I assumed that for V2 at some point we will go through the same escrow path as for the current `RewardsPool`. If both implement the `IRewardsPool` then we will be able to deploy again `CoveragePoolBeneficiary` without changing any code inside. Anyway, it's just an idea, I can hardcode `RewardsPool` without any interfaces.

pdyraga: Nah, that was a good idea though v2 will need a completely different rewarding mechanism for a multi-asset setup.

dimpar: Ok, done in b44262a

contracts/CoveragePoolBeneficiary.sol
dimpar: Sure, done in b44262a

dimpar:
> Will this contract be used in unit/system tests to check if it works?

Yup, tests were added in b44262a

test/CoveragePoolBeneficiary.test.js
```diff
@@ -0,0 +1,74 @@
...
+        await expect(
+          coveragePoolBeneficiary
+            .connect(thirdParty)
+            .__escrowSentTokens(transferAmount)
+        ).to.be.revertedWith("Ownable: caller is not the owner")
+      })
+    })
+
+    context("when a caller is the owner", () => {
+      it("should emit topUpReward event", async () => {
```
pdyraga, line 63: Maybe "should top up reward pool? It is not `__escrowSentTokens` that emits `RewardToppedUp`, it's the rewards pool's `topUpReward` that got called.

dimpar: Done 73b681a

pdyraga (approved):

PR #151, opened 2021-07-07
Handle external contracts in deployment scripts

nkuba:
In this PR we add handling of external contracts covering two types of external artifacts:

1.  Published to NPM registry - used for our contracts migrated in other modules (`keep-core`, `tbtc`) (in both Hardhat or Truffle format).
2. Predefined addresses - used for contracts migrated outside of our org (e.g. `UniswapV2Router`) - we require just an address and optionally an ABI

For more details of changes introduced in this PR please refer to commit messages and updated `README.adoc` file.

TODO:
- [x] update `install.sh` script


external/ropsten/UniswapV2Router.json
```diff
@@ -0,0 +1,90 @@
+{
+  // FIXME: This is just a sample file. Replace it with correct values for given network.
+  "address": "0xZZabcd0000000000000000000000000000000001",
```
nkuba, line 3: Do we know the address of `UniswapV2Router` contract we want to use for ropsten?

README.adoc
```diff
@@ -105,39 +105,138 @@ yarn deploy --network <network>
...
-There is also a possibility to run the deployment with export mode enabled:
+Once the deployment terminates, a new `deployments` directory containing all
+deployment info will be created. It can be directly used by dApps or other client
+code as it contains deployment details like chain ID, transaction hash, ABI or
+address for each contract.
+
+===== Export mode
+
+Apart from deployments saved in the `deployments/` folder the details will be also
+stored in a lightweight file `export.json`. Which contains a handy summary of the
```
michalinacienciala, line 120:
```suggestion
stored in a lightweight file `export.json`, which contains a handy summary of the
```

r-czajkowski (commented):
Want to mention that we also need to update `scripts/install.sh`.  [The parameters] (https://github.com/keep-network/coverage-pools/blob/main/scripts/install.sh#L24-L28) are probably no longer valid.

yarn.lock
```diff
@@ -860,6 +1177,98 @@
...
+    "@openzeppelin/upgrades" "^2.7.2"
+    openzeppelin-solidity "2.4.0"
+
+"@keep-network/keep-core@^1.8.0-dev":
+  version "1.8.0-ropsten.2"
+  resolved "https://registry.yarnpkg.com/@keep-network/keep-core/-/keep-core-1.8.0-ropsten.2.tgz#4b0d4810fa4ae41970cc40c434ab5cc460eca928"
+  integrity sha512-OWuafRUQicjHZImzF0irw0iE+lmxF4wTKyI4kNuht/2O1v/KqRSMM57onUT6rOBRiKVXAX/EjMW0kiJwT2Zdgg==
+  dependencies:
+    "@openzeppelin/upgrades" "^2.7.2"
+    openzeppelin-solidity "2.4.0"
```
michalinacienciala, line 1194: Should we have both `keep-core` dependencies here?

package.json
```diff
@@ -1,5 +1,28 @@
...
+    "lint:fix:sol": "solhint 'contracts/**/*.sol' --fix && prettier --write '**/*.sol'",
+    "lint:js": "eslint . ",
+    "lint:sol": "solhint 'contracts/**/*.sol' && prettier --check '**/*.sol'",
+    "test": "hardhat test",
+    "test:system": "NODE_ENV=system-test hardhat test ./test/system/*.test.js",
+    "test:unit:coverage": "hardhat coverage"
+  },
   "dependencies": {
+    "@keep-network/keep-core": "^1.8.0-dev",
+    "@keep-network/tbtc": "^1.1.2-dev",
```
michalinacienciala, line 25: In tbtc.js/package.json we use "@keep-network/tbtc": ">1.1.2-pre <1.1.2-rc" range, which excludes packages with `-ropsten` suffix. Here we allow for packages with `-ropsten`. Should we? I don't see clear reason why not, but wanted to double check.

scripts/install.sh
```diff
@@ -10,21 +10,13 @@ COVERAGE_POOL_PATH=$(realpath $(dirname $0)/../)
 
 # Defaults, can be overwritten by env variables/input parameters
 NETWORK_DEFAULT="development"
```
dimpar, line 12: how about adding somewhere at the top a verification for a node version? I was on 11.5.0 and it failed on `yarn install` and we can check it at the top. 12.8.1 is just an example that I happen to have, not sure what is the lowest required though
```
NODE_CURRENT_VER="$(node --version)"
NODE_REQUIRED_VER="v12.8.1"

if [ "$(printf '%s\n' "$NODE_REQUIRED_VER" "$NODE_CURRENT_VER" | sort -V | head -n1)" != "$NODE_REQUIRED_VER" ]; 
then
      echo "Required node version must be at least ${NODE_REQUIRED_VER}" 
      exit 1
fi
```

external/ropsten/UniswapV2Router.json
dimpar: I think this is Ropsten's [address](https://ropsten.etherscan.io/address/0x7a250d5630B4cF539739dF2C5dAcb4c659F2488D#code)

package.json
r-czajkowski, line 25: For local development- since we rely on external packages we should link them in the `install.sh` script eg: 
```
yarn link @keep-network/tbtc
yarn link @keep-network/keep-core
```


Note that `tbtc` and `keep-core` use `npm` instead of `yarn` so we have to create links via `yarn link` for `tbtc` and `keep-core`. But I think it should be done in [`keep-core/scripts/start_dashboard.sh`](https://github.com/keep-network/keep-core/blob/main/scripts/start_dashboard.sh)

r-czajkowski (changes requested):

external/ropsten/UniswapV2Router.json
nkuba: Thanks, I updated it in 4645d5e63fe5f106cc2c945cb1a20834f72e5e0f

scripts/install.sh
nkuba: I added `engines` property to `package.json` in e5b4f7c09028c25f99e8d8f477c3ac80d7bdfe35. It should fail the execution:

```
➜  coverage-pools git:(deployment-external-contracts) ✗ yarn install
yarn install v1.17.3
[1/5] 🔍  Validating package.json...
error @keep-network/coverage-pools@0.0.1: The engine "node" is incompatible with this module. Expected version ">= 12.0.0". Got "11.15.0"
error Found incompatible module.
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```

dimpar: nice!

package.json
nkuba: Great catch! 

I added linking in the `install.sh` script. It assumes the links are made before running the script. So we will have to create them in `local-setup` similarly we do for other repos ([example](https://github.com/keep-network/local-setup/blob/5fd3c299943a89d4281ec2021ce863c21015faa6/install-tbtc.sh#L46-L50)).

And for the dashboard exactly as you suggested in `keep-core/scripts/start_dashboard.sh`.

nkuba: Yes, we should be consistent with other repos. We should include just `-dev` packages here. I updated it in 22002144bf07f329609ead9664ec4308f9bb489c.
Please note that for tbtc the range should be `>1.1.2-dev <1.1.2-pre` but the workflow was not yet updated there to publish `dev` packages.

yarn.lock
nkuba: Yes, the dependencies were auto-resolved by yarn, and both are required, please see:
```
➜  coverage-pools git:(deployment-external-contracts) ✗ yarn why @keep-network/keep-core                    
yarn why v1.17.3
[1/4] 🤔  Why do we have the module "@keep-network/keep-core"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
[4/4] 🚡  Calculating file sizes...
=> Found "@keep-network/keep-core@1.8.0-dev.1"
info Has been hoisted to "@keep-network/keep-core"
info This module exists because it's specified in "dependencies".
info Disk size without dependencies: "67.52MB"
info Disk size with unique dependencies: "103.96MB"
info Disk size with transitive dependencies: "266.96MB"
info Number of shared dependencies: 201
=> Found "@keep-network/keep-ecdsa#@keep-network/keep-core@1.8.0-rc.6"
info This module exists because "@keep-network#tbtc#@keep-network#keep-ecdsa" depends on it.
info Disk size without dependencies: "67.55MB"
info Disk size with unique dependencies: "103.98MB"
info Disk size with transitive dependencies: "266.99MB"
info Number of shared dependencies: 201
✨  Done in 2.92s.
```

dimpar (approved):

dimpar (dismissed):

hardhat.config.js
```diff
@@ -32,6 +32,29 @@ module.exports = {
...
+        // deploy: "node_modules/@keep-network/keep-core/deploy",
+      },
+      {
+        artifacts: "node_modules/@keep-network/tbtc/artifacts",
+      },
+    ],
+    deployments: {
+      development: [
+        "../keep-core/solidity/build/contracts",
+        "../tbtc/solidity/build/contracts",
```
r-czajkowski, line 49: Shouldn't we set path to `node-modules`? We link `tbtc` and `keep-core` in `install.sh` (https://github.com/keep-network/coverage-pools/pull/151/commits/7fb710f30ef77dd8a73516afcd28fe9aa46badbd) so I think we should rely on packages.

nkuba: 1d90220c3d5ef87beb2be8ca9847c6a3975c8c47

r-czajkowski (approved):
LGTM! :shipit:

PR #152, opened 2021-07-07
Migrated deployment scripts to TypeScript

nkuba:
In this PR we migrate Hardhat deployments scripts to TypeScript.

Depends on https://github.com/keep-network/coverage-pools/pull/151


deploy/07_deploy_signer_bonds_escrow.ts
```diff
@@ -0,0 +1,18 @@
+import { HardhatRuntimeEnvironment } from "hardhat/types"
+import { DeployFunction } from "hardhat-deploy/types"
+
+const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
+  const { getNamedAccounts, deployments } = hre
+
+  const { deploy } = deployments
```
Shadowfiend, line 7: Fwiw I'm not sure this is any clearer than `hre.deployments.deploy(...)`. Same for the rest of the deconstruction. 

Doesn't need to change, just wanted to call attention for a second look.

nkuba: I agree it may be unnecessary, but it depends on the script's complexness. For example, here we pull out `read` and `write` functions which are not strictly related to the `deployments` but for contracts calls. I'm afraid that `deployments.read` may cause a confusion.
https://github.com/keep-network/coverage-pools/blob/eaa867eb4078eb2fb852045436d506e5eaeac30e/deploy/06_deploy_coverage_pool.ts#L7

I'll update the simple scripts where we deconstruct just the `deploy` function. And I'm waiting for your opinion on the case I mentioned above.

nkuba: I updated the code in 3a8d6796dff73bb6f77bfe8702e60c61f32b833e

deploy/00_resolve_keep_token.ts
```diff
@@ -0,0 +1,28 @@
+import { HardhatRuntimeEnvironment } from "hardhat/types"
+import { DeployFunction } from "hardhat-deploy/types"
+
+const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
+  const { getNamedAccounts, deployments } = hre
```
michalsmiarowski, line 5: My IDE (which is IntelliJ) shows me the warning:

```Property 'getNamedAccounts' does not exist on type 'HardhatRuntimeEnvironment'.```
(and also same for `deployments` property)

It looks like it does not see the properties of `HardhatRuntimeEnvironment` that is placed in `hardhat-deploy` directory but only those placed in `hardhat`. Because of that the code completion in the IDE doesn't work as expected because it doesn't show proper method names after writing (for example) `hre.` or `deployments.`.

The solution for that is adding and empty `tsconfig.json` file to the main directory, although i'm not sure exactly why it works. As far as i know the `tsconfig.json` itself indicates that the directory is the root of a TypeScript project (not talking about the properties inside the file which of course have other purposes that the file itself).

As I can see in the [hardhat documentation](https://google.com) the `tsconfig.json` is optional so i guess it is not necessary to add it.

File: deploy/07_deploy_signer_bonds_escrow.ts [hunk 3]
```diff
@@ -0,0 +1,16 @@
...
+
+  await deployments.deploy("SignerBondsManualSwap", {
+    from: deployer,
+    log: true,
+  })
+}
+
+export default func
+
+func.tags = ["SignerBondsManualSwap"]
```
michalsmiarowski, line 16: Looks like in all other files the tags are the same as the second part of the file name that occurs after `<number>_deploy_`.

So maybe we could change the name of the file here?

`07_deploy_signer_bonds_escrow.ts` -> `07_deploy_signer_bonds_manual_swap.ts`

deploy/10_deploy_risk_manager_v1.ts
```diff
@@ -1,6 +1,11 @@
-module.exports = async ({ getNamedAccounts, deployments }) => {
```
michalsmiarowski, line 1: Can't add comment to the proper lines, because there were no changes in them so I add this here, on top of the file.

Lines 16-19
```
const auctionLength = 86400 // 24h
  const bondAuctionThreshold = 100
  const initialSwapStrategy =
    process.env.INITIAL_SWAP_STRATEGY || "SignerBondsManualSwap"
```

I think since we use TypeScript here we should get used to declaring the types for the variables that we declare even if we are setting the value for them right away (see examples in the [documentation](https://www.typescriptlang.org/docs/handbook/basic-types.html).

So i would change this code to:
```
const auctionLength: number: number = 86400 // 24h
  const bondAuctionThreshold: number = 100
  const initialSwapStrategy: string =
    process.env.INITIAL_SWAP_STRATEGY || "SignerBondsManualSwap"
```

I think we do not need to specify the types for the variables that comes from `hre` and its properties, grand-properties etc. (although we probably if you think it will improve readability) since `HardhatRuntimeEnvironment` take care of that (but please see my first comment about `tsconfig.json`)

Thoughts?

michalsmiarowski: I think that we could also specify the units for `bondAuctionThreshold` in the comment. For `auctionLength` it is clear that the unit is in seconds since we have information that it's 24h. For `bondAuctionThreshold` it is not so clear IMO.

michalsmiarowski (changes requested):
Overall looks brilliant! 💎 

Left some comments to look into before the merge.

deploy/00_resolve_keep_token.ts
r-czajkowski: VSC shows me the same warning. Adding the `tsconfig.json` did the trick. I think we should take a look at [typescript support for hardhat-deploy plugin](https://hardhat.org/plugins/hardhat-deploy.html#typescript-support).

deploy/07_deploy_signer_bonds_escrow.ts
r-czajkowski: I think it makes seanse. The `SignerBondsEscrow` is now called `SignerBondsManualSwap`- please see 874455749efe522f06e634c8e36ef977a4a3cb44.

deploy/00_resolve_keep_token.ts
nkuba: 9c206b3459accc8b29a668d24393656f5c2bbfe4

deploy/07_deploy_signer_bonds_escrow.ts
nkuba: 731808a1f3216994794a6b1ee0750f342bce7b07

deploy/10_deploy_risk_manager_v1.ts
nkuba: fdde04b83c6a6d6c8f0e12af8a65240d645cd6d1

michalsmiarowski (approved):
LGTM 👍

r-czajkowski (approved):
LGTM!

PR #153, opened 2021-07-08
Follow-up for deployment script

nkuba:
We want to compare addresses securely by converting them to consistently formatted addresses.

We renamed the default network from `local` to `development` and missed to update `install.sh` script.

### DIFF
--- a/deploy/06_deploy_coverage_pool.js
+++ b/deploy/06_deploy_coverage_pool.js
@@ -9,7 +9,10 @@ module.exports = async ({ getNamedAccounts, deployments }) => {
     log: true,
   })
 
-  if ((await read("AssetPool", "owner")) !== CoveragePool.address) {
+  if (
+    ethers.utils.getAddress(await read("AssetPool", "owner")) !==
+    ethers.utils.getAddress(CoveragePool.address)
+  ) {
     log(`transferring ownership of AssetPool to ${CoveragePool.address}`)
 
     await execute(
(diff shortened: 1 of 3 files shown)


deploy/06_deploy_coverage_pool.js
```diff
@@ -9,7 +9,10 @@ module.exports = async ({ getNamedAccounts, deployments }) => {
     log: true,
   })
 
-  if ((await read("AssetPool", "owner")) !== CoveragePool.address) {
+  if (
+    ethers.utils.getAddress(await read("AssetPool", "owner")) !==
+    ethers.utils.getAddress(CoveragePool.address)
```
r-czajkowski, line 14: Seems like it can be reusable in other scripts, so maybe we can extract to a separate function or even add to the [hre object](https://hardhat.org/advanced/hardhat-runtime-environment.html#extending-the-hre). Ofc non-blocking comment!

```
function isSameAddress(address1, address2) {
 ethers.utils.getAddress(address1) !==
 ethers.utils.getAddress(address2)
}
```

r-czajkowski (approved):
LGTM! Left a non-blocking note for consideration.

deploy/06_deploy_coverage_pool.js
nkuba: Good idea, I'll handle it in a follow-up PR.

nkuba: Covered in https://github.com/keep-network/coverage-pools/pull/162

PR #158, opened 2021-07-12
Collateral token described in docs

### LINKED ISSUE #35: Reentrancies risks on Auction.takeOffer
Severity: Undetermined

## Description

`Auction.takeOffer` calls `transferFrom` on an external token without having reeentrancy protection:

https://github.com/keep-network/coverage-pools/blob/49ba7e9dfb5fb421c2b5db18496479d25d374072/contracts/Auction.sol#L146-L157

If the token used has a callback mechanism (such as an ERC777), the destination can re-enter into the contract. 

We did not find a direct way to profit from this, but we considered:
- Re-entering into `takeOffer`: this will lead `_onOffer` to be computed twice with the same value; however we did not find a way for an attacker to profit from it
- Re-entering to execute `earlyClose`, which could lead the system to call `harikari` multiple times. However, this requires privileged accesses, and it is unclear how an attacker would make a profit

## Recommendations
Short term, investigate the risks of reentrancies on tokens transfer. Consider performing the token transfer at the end of the function, or adding reentrancy guard to `takeOffer`. Additionally, take into consideration the reentrancy risks on token transfers when designing `CollateralPool.seizeFunds`.

Long term, be aware of the risks associated with arbitrary tokens, and review the [token integration checklist](https://github.com/crytic/building-secure-contracts/blob/master/development-guidelines/token_integration.md).

### LINKED ISSUE #137: [Fix review] receiveApproval might still be vulnerable
See https://github.com/keep-network/coverage-pools/issues/59

The fix proposed was: 
https://github.com/keep-network/coverage-pools/blob/09aa12565ab46a954ab2cb969209e6721bd6b631/contracts/AssetPool.sol#L137

While the original fix looked ok, after more consideration, there is still a risk for a token allowing an **arbitrary data** execution on transfers/approvals. Those are less likely to be used in the system, but might still happen.

One way to mitigate this is to manually review the `token` used, to ensure it does not allow such behavior. This must be documented and checked at deployment.

pdyraga:
Closes #137 
Closes #35

Described in v1 documentation what is a collateral token. Also, made it
explicit collateral token can not be an arbitrary asset but needs to be
a trusted ERC20 that passed a security checklist and does not impose
any security thread. Coverage pool code is not secure against evil tokens
with, for example, reentrant malicious code.

### DIFF
--- a/docs/design.adoc
+++ b/docs/design.adoc
@@ -13,6 +13,15 @@ covered in v2 documentation.
 
 == Components
 
+=== Collateral token
+
+ERC-20 token deposited in the pool by underwriters and accepted by the pool as
+collateral. Collateral token positions of underwriters can be affected when the
+Risk manager claims coverage. Token accepted as collateral needs to pass
+https://github.com/crytic/building-secure-contracts/blob/master/development-guidelines/token_integration.md[token security checklist]
+and should not impose any security risk on the coverage pool like a possible
+re-entrancy or allowing an arbitrary data execution on transfers/approvals.
+
 === Asset pool
 
 Asset pool accepts a single ERC-20 token as collateral, and returns an
--- a/docs/design.adoc
+++ b/docs/design.adoc
@@ -15,9 +15,9 @@ covered in v2 documentation.
 
 === Collateral token
 
-ERC-20 token deposited in the pool by underwriters and accepted by the pool as
-collateral. Collateral token positions of underwriters can be affected when the
-Risk manager claims coverage. Token accepted as collateral needs to pass
+ERC-20 token which is accepted by the pool as collateral and used as a deposit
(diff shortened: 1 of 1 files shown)


docs/design.adoc
```diff
@@ -13,6 +13,15 @@ covered in v2 documentation.
 
 == Components
 
+=== Collateral token
+
+ERC-20 token deposited in the pool by underwriters and accepted by the pool as
```
dimpar, line 18: I would change the order, ie. first mention that it is accepted by the pool and then these tokens can be accepted as a deposit. How about something like that:
"ERC-20 token which is accepted by the pool as collateral and used as a deposit token by underwriters."

pdyraga: Yeah, makes sense! I'll update.

dimpar (approved):

PR #162, opened 2021-07-13
Use @keep-network/hardhat-helpers

nkuba:
The plugin contains useful functions we will use in deployment scripts.

I also noticed that we need to add an additional validation for address obtained from an artifact as it happens that the address is undefined (0f3a443).


r-czajkowski (approved):
Brilliant idea with a custom plugin and +1 for separate repo for `@keep-network/hardhat-helpers`. :shipit:

PR #155, opened 2021-07-08
Add a job to deploy contracts on testnet

michalinacienciala:
This commit/PR adds a deployment of coverage-pools contracts on a
testnet (name of the testnet on which deployment will take place will
depend on the value of `environment` input passed when invoking
`workflow_dispatch` event). Before deploy script will run, we'll update
dependencies to upstream packages (`keep-core` and `tbtc` contracts are
needed for the deployment of `coverage-pools` contracts.)
Once contracts get deployed, we will push them to Tenderly (only
`ropsten` env curently supported) and we will publish npm package (which
will be later pulled by the Token Dashboard - it's not the scope of this
PR).
Last step in the job is an execution of `notify-workflow-completed`
action which notifies CI about successful run of the workflow and as a
result triggers deploy of any downstream modules specified in the
`ci/onfig.config.json` file.
As part of this PR we also change the conditions of execution of
`contracts-lint` and `contracts-slither` jobs - those jobs will no
longer be executed for `workflow_dispatch` event.

Depends on:
https://github.com/keep-network/coverage-pools/pull/165
https://github.com/keep-network/coverage-pools/pull/168

TODO:

- [x] Remove `--dry-run` in publish step before merging
- [x] Configure an access to `TENDERLY_TOKEN` from `coverage-pools` repo

PR #163, opened 2021-07-15
Update references to hardhat config after switching from JS to TS

michalinacienciala:
We're no longer storing the hardhat config in JavaScript format. Instead
we're using TypeScript. We need to change references to the config from
`hardhat.config.js` to `hardhat.config.ts`.

### DIFF
--- a/.github/workflows/contracts.yml
+++ b/.github/workflows/contracts.yml
@@ -80,7 +80,7 @@ jobs:
 
       - name: Install Solidity
         env:
-          SOLC_VERSION: 0.7.6 # according to solidity.version in hardhat.config.js
+          SOLC_VERSION: 0.7.6 # according to solidity.version in hardhat.config.ts
         run: |
           pip3 install solc-select
           solc-select install $SOLC_VERSION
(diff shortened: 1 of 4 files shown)


.github/workflows/contracts.yml
```diff
@@ -80,7 +80,7 @@ jobs:
 
       - name: Install Solidity
         env:
-          SOLC_VERSION: 0.7.6 # according to solidity.version in hardhat.config.js
+          SOLC_VERSION: 0.7.6 # according to solidity.version in hardhat.config.ts
```
michalinacienciala, line 83: This may not necessary be in scope of this PR, but I noticed that the version here doesn't match what's in the `hardhat.config.ts` (`0.8.4` is set in there). Should we update the version here or maybe get rid of the comment?

nkuba: Good catch! We should update it to match the version from the file.

nkuba (approved):

.github/workflows/contracts.yml
michalinacienciala: Done as part of https://github.com/keep-network/coverage-pools/pull/155.

Issue #125, opened 2021-06-27
Coverage Pool v1.0.0 deployment


pdyraga:
Team | Members
------------ | -------------
**Deployment Team** | `0x123694886DBf5Ac94DDA07135349534536D14cAf`
**Technical Wallet Team** | `0xB3726E69Da808A689F2607939a2D9E958724FC2A`
**Community Multisig** | `0x19FcB32347ff4656E4E6746b4584192D185d640d`
**Reward Manager** | `0xb78c0cf4c9e9bf4ba24b17065fa8c0ac71957653`
**Communications** | TBD

# Prerequisites
- [x] Determine initial auction length (https://github.com/keep-network/coverage-pools/issues/126)
- [x] Deployment scripts ready
- [x] All security audit issues addressed

# Deploy smart contracts
All steps done by Deployment Team
- [x] Tag `solidity/v1.0.0` in `keep-network/coverage-pools`
- [x] Sign `solidity/v1.0.0` commit hash
- [x] Deploy `UnderwriterToken` contract (address: `0x2Aa24DAc5E494e7B028ED43023530E5769df5D8b`) with the following parameters:
  - `_name`: `covKEEP underwriter token`
  - `_symbol`: `covKEEP`
- [x] Deploy `AssetPool` (address: `0x2ada2d985Ce69b467d2D8372494Df021BEDfdbf2`) with the following parameters:
  - `_collateralToken`: `KEEP@0x85eee30c52b0b379b046fb0f85f4f3dc3009afec`
  - `_underwrtierToken`: `UnderwriterToken.address` (deployed in previous step)
  - `rewardsManager`: Deployment Team (temporarily, will be transferred in next steps)
- [x] Deploy `CoveragePool` (address: `0x91cd77a78de10e9df7538270b324a8a8be3bf877`) with the following parameters:
  - `_assetPool`: `AssetPool.address` (deployed in previous step)
- [x] Deploy `AuctionBidder` (address: `0x773A4cACbcA8e3704F8f7842531eAb83e437B3C4`) with the following parameters:
  - `_coveragePool`: `CoveragePool.address` (deployed in previous step)
- [x] Deploy `SignerBondsManualSwap` (address: `0x43faf6fe84e01dd186cfeb9ea2f0854a1ef7442f`)
- [x] Deploy `SignerBondsUniswapV2`  (address: `0x9f054B7EaEEE0011A3bF3eAe38E73fb7A37e1eAe`) with the following parameters:
[...]

PR #165, opened 2021-07-20
Mainnet deployment scripts

nkuba:
This PR updates deployment scripts to reflect mainnet specific scenario described in https://github.com/keep-network/coverage-pools/issues/125.

It also adds the possibility to configure secret network details in a file that will not be committed.


hardhat.config.ts
```diff
@@ -62,14 +75,21 @@ const config: HardhatUserConfig = {
         "node_modules/@keep-network/tbtc/artifacts",
         "./external/ropsten",
       ],
+      mainnet: ["./external/mainnet"],
```
nkuba, line 78: This solution assumes the artifacts from `keep-core` and `tbtc` that we need addresses for are committed to the repository. I decided to go with it to reduce the possibility of using wrong addresses if we use npm packages.

An alternative way would require executing `yarn add @keep-network/keep-core@1.7.0 @keep-network/tbtc@1.1.0` to install packages from specific mainnet versions and update this line to:
```ts
      mainnet: [
        "node_modules/@keep-network/keep-core/artifacts",
        "node_modules/@keep-network/tbtc/artifacts",
        "./external/mainnet"
      ],
```

hardhat.config.ts
```diff
@@ -62,14 +75,21 @@ const config: HardhatUserConfig = {
...
   },
   namedAccounts: {
     deployer: {
       default: 0, // take the first account as deployer
+      ropsten: RopstenSecrets.address,
+      mainnet: MainnetSecrets.address,
     },
     rewardManager: {
       default: 1,
+      mainnet: "0xB3726E69Da808A689F2607939a2D9E958724FC2A", // Technical Wallet // TODO: Technical wallet or rewards manager?
```
nkuba, line 89: @pdyraga would you please double-check this https://github.com/keep-network/coverage-pools/issues/125#issuecomment-882472963.

pdyraga: Let's use deployer address, this role is later transferred anyway.

hardhat.config.ts
pdyraga: I think the approach you used is better. We need them only for the deployment. Does it mean we can remove them from `package.json`? 
```
    "@keep-network/keep-core": ">1.8.0-dev <1.8.0-pre",
    "@keep-network/tbtc": ">1.1.2-dev <1.1.2-ropsten",
```

pdyraga: Ah, no, we can't because they are used for `development` network earlier, right? So ones in `package.json` should always stay dev-only. If that's correct, can we drop a comment in `package.json` explaining these dependencies - why they are needed, where they are referenced, and what versions (dev-only) are expected?

hardhat.config.ts
pdyraga: This still has to be reworked, right?

deploy/13_deploy_coverage_pool_beneficiary.ts
```diff
@@ -0,0 +1,39 @@
+import { HardhatRuntimeEnvironment } from "hardhat/types"
```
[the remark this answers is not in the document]
nkuba: I may be missing some context, but these two are not related, right? Will the ordering improve something? I like the current order as we keep mainnet-only scripts as the last ones.

hardhat.config.ts
nkuba: They are used for development and will be used for ropsten. Basically, we need them in `package.json` for any network that artifacts will be changing dynamically. For `ropsten` we will be updating the package version to a specific one deployed on ropsten.

Unfortunately, we cannot add comments in a JSON file :( But this approach is a general approach that we have in all our repositories.

hardhat.config.ts
nkuba: Done now

deploy/06_deploy_coverage_pool.ts
```diff
@@ -14,40 +13,20 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
...
+    "AssetPool",
+    CoveragePool.address,
+    deployer
+  )
+
+  await helpers.ownable.transferOwnership(
+    "UnderwriterToken",
+    AssetPool.address,
+    deployer
+  )
```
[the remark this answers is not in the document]
nkuba: > In this context, I do not see where do we transfer RewardsPool ownership to CoveragePoolBeneficiary 🤔

198b5e164f343d3fa2e7d32e45ae1ee989b90ca9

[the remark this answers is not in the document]
nkuba: > Wouldn't it be better to have all ownership transfers aggregated in a separate step, just like we have for community multisig transfer in step 14? Less chance we'll miss something.

As we separated scripts for contracts to separate files to easier track them and make use of tags and dependencies I believe it's better to keep the ownership transfers in the individual places rather than gathering them all together.

deploy/13_deploy_coverage_pool_beneficiary.ts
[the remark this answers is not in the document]
pdyraga: I was looking at it as: deploy everything first, then configure. But after your explanation that mainnet-only steps are executed as last ones, the current order makes total sense 👍

deploy/06_deploy_coverage_pool.ts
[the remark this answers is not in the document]
pdyraga: In that case, shouldn't we have
https://github.com/keep-network/coverage-pools/blob/063db74969fed5c6ff75455d4487ba2150f424f4/deploy/06_deploy_coverage_pool.ts#L22-L27

in `05_deploy_asset_pool.ts` instead of in `06_deploy_coverage_pool.ts`?

pdyraga:
Not strictly related to code changes in this PR but if we consider this PR as the final one for deployment, I'd like to address one more issue. [Here](https://github.com/keep-network/coverage-pools/blob/87a5f22c435648680b96b6e9f20c17ea1d1933fc/deploy/09_deploy_master_auction.ts#L8) we deploy the `Auction` contract under the name `MasterAuction`, I take it's how it will be represented in the artifact we publish to NPM. In other words, everyone wanting to work with `Auction` contract will have to instantiate `MasterAuction` instance. Is it true? If so, why not just `Auction`? I do understand we deploy a *master* instance and we can call it like that in the deployment script, but if we require everyone to instantiate master auction clones using `MasterAuction` name, it's wrong - the contract name is `Auction`. Maybe I missed some context on how the final NPM artifacts are assembled based on the name used in deployment scripts. If they will be instantiating `Auction`, we are fine.

deploy/06_deploy_coverage_pool.ts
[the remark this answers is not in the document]
nkuba: d19606cf8fb666d27c1fd9be2483f9cf54007839

nkuba:
> Not strictly related to code changes in this PR but if we consider this PR as the final one for deployment, I'd like to address one more issue. [Here](https://github.com/keep-network/coverage-pools/blob/87a5f22c435648680b96b6e9f20c17ea1d1933fc/deploy/09_deploy_master_auction.ts#L8) we deploy the `Auction` contract under the name `MasterAuction`, I take it's how it will be represented in the artifact we publish to NPM. In other words, everyone wanting to work with `Auction` contract will have to instantiate `MasterAuction` instance. Is it true? If so, why not just `Auction`? I do understand we deploy a _master_ instance and we can call it like that in the deployment script, but if we require everyone to instantiate master auction clones using `MasterAuction` name, it's wrong - the contract name is `Auction`. Maybe I missed some context on how the final NPM artifacts are assembled based on the name used in deployment scripts. If they will be instantiating `Auction`, we are fine.

Hardhat clearly separates deployment artifacts from build artifacts. Deployments artifacts are the ones that got deployed and have an address and deployment details, these will be stored in `artifacts/` (`artifacts/MasterAuction.json`).
Build artifacts contain just compilation data like abi, bytecode, etc. These are stored in `build/contracts/` directory (`build/contracts/Auction.sol/Auction.json`).

The build artifacts can be easily used with [`hardhat`](https://hardhat.org/guides/compile-contracts.html#reading-artifacts) (`hre.artifacts.readArtifact(“Auction")`) and [`hardhat-ethers`](https://github.com/nomiclabs/hardhat/tree/master/packages/hardhat-ethers#helpers) (`ethers.getContractAt(“Auction”, “0x00..01”)`).

Only if users want to use other solutions like web3 they would have to read the abi from an artifact in one of two places:
- `artifacts/MasterAuction.json`
- `build/contracts/Auction.sol/Auction.json`

I can see the advantages and disadvantages of the current naming and the one you proposed. Whatever we decide to do it won't be perfect.

I pushed a change in 21b7dc4a5b93454c2640a626782f6186c6c8b317. But if after my explanation you change your mind I'll be happy to revert it.

pdyraga (approved):

PR #166, opened 2021-07-22
Solidity version cleanup

pdyraga:
Two small cleanup changes:

- Use 0.8.4 solc version for contracts workflow: we changed the solidity version in all files to 0.8.4, th same version is referenced in hardhat.config.js
- 0.8.4 solidity pragma used for GovernanceUtils: This was the last file in which the solidity version was not pinned to
0.8.4. Looks that we mistakenly omitted it.

### DIFF
--- a/contracts/GovernanceUtils.sol
+++ b/contracts/GovernanceUtils.sol
@@ -12,7 +12,7 @@
 
 // SPDX-License-Identifier: MIT
 
-pragma solidity <0.9.0;
+pragma solidity 0.8.4;
 
 library GovernanceUtils {
     /// @notice Gets the time remaining until the governable parameter update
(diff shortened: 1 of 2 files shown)


contracts/GovernanceUtils.sol
```diff
@@ -12,7 +12,7 @@
 
 // SPDX-License-Identifier: MIT
 
-pragma solidity <0.9.0;
+pragma solidity 0.8.4;
```
nkuba, line 15: Why we don't do semver `^0.8.4`?

pdyraga: We do not want to stay on the most common one, per the recommendation of auditors.

nkuba (approved):

Issue #159, opened 2021-07-13
Define NPM package structure for deployed artifacts


pdyraga:
Hardhat and Truffle use a different package structure. We need to decide how our NPM packages should look like so that they can be used from Dashboard. 

Refs: CP-20

nkuba:
For our other projects based on **Truffle**, we export compilation and deployment artifacts in truffle JSON files located in `./artifacts/` directory. It is a symlink to `./build/contracts` which is the default path where Truffle places the artifacts.
The directory structure is flat, artifacts are placed directly under the `./artifacts/` directory (e.g. `./artifacts/KeepToken.json`). The file contains `abi` property that holds ABI generated from contract source code, and deployment details that are held by `network` property with mapping to network ID, e.g.:
```json
{
  "abi": [
    {
      "inputs": [
        {
          "internalType": "address",
          "name": "TestToken",
          "type": "address"
        }
      ],
      "payable": false,
      "stateMutability": "nonpayable",
      "type": "constructor"
    }
  ],
  "networks": {
    "1101": {
      "events": {},
      "links": {},
      "address": "0x78fecE70FEF77301006e74146B1d823ec7BfC179",
      "transactionHash": "0xdb5da0e5171ce0dd7fd20ddb02fd921fa2863507d6da586401a24390cbc115fb"
    }
  }
}
```
# Hardhat

**Hardhat** with its' plugin [hardhat-deploy](https://github.com/wighawag/hardhat-deploy) produces artifacts in a different way.
## Build artifacts

 Compilation artifacts are separated from deployment artifacts. Compilation artifact goes to a JSON file under `./artifacts/contracts/` directory reflecting the source code directory structure, e.g. `./artifacts/contracts/v2/RewardsPoolV2.sol/RewardPoolV2.json`, the file contains ABI.

<details>
  <summary>Sample Hardhat compilation artifact</summary>

```json
{
  "_format": "hh-sol-artifact-1",
  "contractName": "RewardPoolV2",
  "sourceName": "contracts/v2/RewardsPoolV2.sol",
  "abi": [
    {
      "inputs": [
        {
          "internalType": "address",
          "name": "assetPool",
          "type": "address"
        }
      ],
      "name": "earned",
      "outputs": [
        {
          "internalType": "uint256",
          "name": "",
          "type": "uint256"
        }
      ],
      "stateMutability": "view",
      "type": "function"
    }
  ],
  "bytecode": "0x1111",
[...]

PR #168, opened 2021-07-23
Prepare NPM package structure for package publication

nkuba:
This PR implements what was previously described in https://github.com/keep-network/coverage-pools/issues/159.

## Structure

The package will have the following structure:
```
artifacts/               # copy of deployments/<network>/
build/contracts/         # build artifacts
contracts/               # source code
export.json              # single-file deployment export
```

- `artifacts/` - is a directory containing deployed contracts artifacts for the network the package is being created. It is assumed that a package is published for each supported network separately as per the [RFC-18](https://github.com/keep-network/keep-core/blob/main/docs/rfc/rfc-18-release-management.adoc), so we don't need to create separate directories per network.
- `build/contracts/` - is a directory containing compiled contracts artifacts in the Hardhat format.
- `contracts/` - is a directory containing contracts source code.
- `export.json` - lightweight deployment data in a single file.

## Scripts

There was a script created to prepare the `artifacts/` directory. The script can be executed with:
```
./scripts/prepare-artifacts.sh --network <network>
```

In `package.json` we defined a `prepublishOnly` entry that will execute before the package is packed and published if `npm publish` command is run. It expects the network option to be provided like in the following example:
```
npm publish --dry-run --network=ropsten
```
or
```
npm_config_network=development yarn publish
```

The command can be also executed separately with:
```
npm run prepublishOnly --network=ropsten
```
or
```
npm_config_network=development yarn prepublishOnly
```

Closes: https://github.com/keep-network/coverage-pools/issues/159


nkuba:
@michalinacienciala We can use this feature in CI to prepare the package for publication after migration.

michalinacienciala:
> @michalinacienciala We can use this feature in CI to prepare the package for publication after contracts migration.
Applied in https://github.com/keep-network/coverage-pools/pull/155.

michalinacienciala:
It may not be strictly related to changes in this PR, but maybe we could update [this line in README.adoc](https://github.com/keep-network/coverage-pools/blame/npm-package-structure/README.adoc#L90) to "Compiled contracts will land in the `build/contracts` directory"?

michalinacienciala (commented):
I've tested the script locally and haven't found any issues. The only comment I had was not strictly related to changes in this PR. Leaving for @r-czajkowski to review.

r-czajkowski (commented):
I think we should run `prepare-artifacts.sh` script instead of creating a symlink in the [`install.sh`](https://github.com/keep-network/coverage-pools/blob/main/scripts/install.sh#L69). Ofc non-blocking comment and we can address it in a separate PR.

Looking good on my side! 👍

nkuba:
Nice findings @michalinacienciala and @r-czajkowski! I updated the code, ready for a final shot!

r-czajkowski (approved):
LGTM! 👍

michalinacienciala (approved):

PR #173, opened 2021-08-19
Cancelling already approved new asset pool

dimpar:
In case of some misconfiguration of a new asset pool, we give a governance an option to set already approved asset pool back to zero address.

### DIFF
--- a/test/AssetPool.test.js
+++ b/test/AssetPool.test.js
@@ -5,6 +5,7 @@ const {
   lastBlockTime,
   to1ePrecision,
   pastEvents,
+  ZERO_ADDRESS,
 } = require("./helpers/contract-test-helpers")
 
 describe("AssetPool", () => {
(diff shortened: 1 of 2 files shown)


test/AssetPool.test.js
```diff
@@ -1190,6 +1191,60 @@ describe("AssetPool", () => {
...
+      await newUnderwriterToken.deployed()
+
+      const NewAssetPoolStub = await ethers.getContractFactory(
+        "NewAssetPoolStub"
+      )
+      newAssetPool = await NewAssetPoolStub.deploy(
+        collateralToken.address,
+        newUnderwriterToken.address
+      )
+      await newAssetPool.deployed()
```
tomaszslabon, line 1214: Could we move objects related to the new asset pool from `cancelNewPoolUpgrade` and `approveNewAssetPoolUpgrade` to the [main section of AssetPool.test.js](https://github.com/keep-network/coverage-pools/blob/951045379ce38f5e73fd61991f84a946a18b62fc/test/AssetPool.test.js#L25)? They are duplicated in both tests.

pdyraga: I'd leave them as-is. It's true they are duplicated but given that they are duplicated only one time, in the context of `approveNewAssetPoolUpgrade` and the rest of the tests in this file do not need them, I think it's better to keep this code as-is.

pdyraga (approved):
LGTM, leaving the final decision to @tomaszslabon

test/AssetPool.test.js
tomaszslabon: @pdyraga I'm ok with leaving it as is.

PR #174, opened 2021-08-19
Removed redundant balance checks, simplified TestToken

pdyraga:
The main change in this PR is the removal of redundant `balanceOf` 
checks in the  `AssetPool`.

In case of `initiateWithdrawal`, `safeTransferFrom` will fail in case
the withdrawing party has not enough covKEEP on their balance. covKEEP
inherits from `ERC20WithPermit` and behaves correctly so there is no risk
here.

In case of `upgradeToNewAssetPool`, `burnFrom` in the underwriter token
fails in case the upgrading party has not enough on their balance.
covKEEP inherits from `ERC20WithPermit` and behaves correctly so there is
no risk here.
To make the revert cause in this case more clear, I have switched the
order of `burnFrom` and `depositFor` calls. In case there is not enough
tokens on underwriter's balance, it is better to revert saying burn
amount exceeds balance instead of reverting with transfer amount exceeds
balance in the new asset pool, in case the new underwriter exceeded the
amount so significantly that the transfer amount exceeds pool's balance.

I have packed two other small changes into this PR:
- Removed unused import to `AssetPool` from `UnderwriterToken` contract.
- Made `TestToken` to inherit from `ERC20WithPermit` to reduce the complexity
in the test contract to an absolute minimum.

### DIFF
--- a/contracts/AssetPool.sol
+++ b/contracts/AssetPool.sol
@@ -212,12 +212,6 @@ contract AssetPool is Ownable, IAssetPool {
     /// @dev Before calling this function, underwriter token needs to have the
     ///      required amount accepted to transfer to the asset pool.
     function initiateWithdrawal(uint256 covAmount) external override {
-        uint256 covBalance = underwriterToken.balanceOf(msg.sender);
-        require(
-            covAmount <= covBalance,
-            "Underwriter token amount exceeds balance"
-        );
-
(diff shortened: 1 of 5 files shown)


dimpar:
Reviewed contracts only, so far looks good.

dimpar (approved):

PR #175, opened 2021-08-20
Replacing storage velocity depleting rate with in memory variable

dimpar:
We can simplify calculation a bit and save on some gas by moving calculation of `velocityPoolDepletingRate` to the `_onOffer()` function.


pdyraga:
@dimpar can you please rebase to squeeze these commits into one?

dimpar:
> @dimpar can you please rebase to squeeze these commits into one?

Sure, done!

pdyraga (approved):

PR #177, opened 2021-08-20
Check cloned auction address

tomaszslabon:
This PR adds:
 - **a check that reverts if the address returned from `CloneFactory` is zero**
 Note: The added check is not strictly necessary, as shortly after the auction creation we call `initialize` on the new auction contract and it will cause the transaction to revert if the cloned contract address is `0`.
 I couldn't find any info on whether the `create` opcode can return 0 or not.
 I think it could happen if the  master auction was destroy as stated [in the original documentation](https://github.com/optionality/clone-factory):
 
> Do not allow your master contract to be self-destructed as it will cause all clones to stop working, thus freezing their state and balances.
 
  
- **replaces lokal version of `CloneFactory` with the one from `thesis/solidity-contracts`**
  Note: it was necessary to update the solidity version from `0.8.4` to `0.8.5` in all the contracts as the `CloneFactory` from `thesis/solidity-contracts` requires `solidity ^0.8.5`.


pdyraga:
Let's also use `CloneFactory` from https://github.com/thesis/solidity-contracts/blob/main/contracts/clone/CloneFactory.sol

pdyraga:
Let's also pin the version in `TBTCDepositTokenStub.sol`. Otherwise, I am good to go.

pdyraga:
I have also opened https://github.com/thesis/solidity-contracts/pull/11 to emphasize in `CloneFactory` docs that 0x0 needs to be handled. I suggest we merge PR 11 first, and then update `package.json` in this PR to point to the most recent commit hash of `thesis/solidity-contracts`. This will let us verify contracts on Etherscan with this change in docs upon the deployment.

tomaszslabon:
> Let's also pin the version in `TBTCDepositTokenStub.sol`. Otherwise, I am good to go.

Done in https://github.com/keep-network/coverage-pools/commit/28bd524a53380aa0c8f0c66282dee99923a5316e.

tomaszslabon:
> I have also opened [thesis/solidity-contracts#11](https://github.com/thesis/solidity-contracts/pull/11) to emphasize in `CloneFactory` docs that 0x0 needs to be handled. I suggest we merge PR 11 first, and then update `package.json` in this PR to point to the most recent commit hash of `thesis/solidity-contracts`. This will let us verify contracts on Etherscan with this change in docs upon the deployment.

I merged PR11 and updated commit hash of `solidity-contracts` in this PR to the latest one in https://github.com/keep-network/coverage-pools/commit/70e52e5e1496b140ad6fa84cd60f4f99929e9771.

package.json
```diff
@@ -28,7 +28,7 @@
   "dependencies": {
     "@keep-network/keep-core": ">1.8.0-dev <1.8.0-pre",
     "@keep-network/tbtc": ">1.1.2-dev <1.1.2-ropsten",
-    "@thesis/solidity-contracts": "github:thesis/solidity-contracts#bf514ba",
+    "@thesis/solidity-contracts": "github:thesis/solidity-contracts#a1384af",
```
pdyraga, line 31: Let's update `yarn.lock` as well.

tomaszslabon: Done in https://github.com/keep-network/coverage-pools/commit/e8bed549b7143437a0a70d19b3ba5cc82b2687ba.

pdyraga (approved):

PR #178, opened 2021-08-23
Update hardhat version and fix unit tests

tomaszslabon:
Due to an [error](https://github.com/nomiclabs/hardhat/issues/1415) in Hardhat `2.2.0`, we decided to update Hardhat version and fix errors in unit and system tests.

The errors in tests were caused by functions like `isOpen`, `takeOffer`, `earlyClose` being called on auctions that self destructed (the previously used version of Hardhat seemed to tolerate such calls).

This PR contains the following changes:
- updates hardhat requirement from `^2.1.1` to `^2.2.1` (it seems `npm` installed the newest version -  `2.6.1`)
- adds helper function `isCodeAt(address)` which checks whether an address stores contract  code, which seems to be [the best way to check if contract self-destructed](https://ethereum.stackexchange.com/a/83024)
- uses `isCodeAt` instead of calling `isOpen` for checking auction self-destruction or relying on reverts

Note: Unfortunately, in two system tests a significant increase of gas usage was observed.
Note: The contracts themselves were not modified by this PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "eslint-config-keep": "github:keep-network/eslint-config-keep#0c27ade",
     "ethereum-waffle": "^3.3.0",
     "ethers": "^5.0.32",
-    "hardhat": "^2.2.1",
+    "hardhat": "2.6.1",
     "hardhat-deploy": "^0.8.11",
     "hardhat-gas-reporter": "^1.0.4",
     "prettier": "^2.3.2",
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "eslint-config-keep": "github:keep-network/eslint-config-keep#0c27ade",
     "ethereum-waffle": "^3.3.0",
     "ethers": "^5.0.32",
-    "hardhat": "2.6.1",
+    "hardhat": "^2.6.1",
     "hardhat-deploy": "^0.8.11",
     "hardhat-gas-reporter": "^1.0.4",
     "prettier": "^2.3.2",
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "eslint-config-keep": "github:keep-network/eslint-config-keep#0c27ade",
     "ethereum-waffle": "^3.3.0",
     "ethers": "^5.0.32",
-    "hardhat": "^2.1.1",
+    "hardhat": "^2.2.1",
     "hardhat-deploy": "^0.8.11",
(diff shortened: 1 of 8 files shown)


test/Auction.test.js
```diff
@@ -163,6 +167,10 @@ describe("Auction", () => {
         await increaseTime(24000)
       })
 
+      it("should not self destruct", async () => {
+        expect(await isCodeAt(auction.address)).to.be.true
```
pdyraga, line 171: We are testing `onOffer` view function and this function can not self-destruct. I'd move this check to "should be opened" test, just like we have in the previous context.

test/system/liquidation.test.js
```diff
@@ -136,7 +137,7 @@ describeFn("System -- liquidation", () => {
     })
 
     it("should close auction", async () => {
```
pdyraga, line 139: "should destroy auction"?

test/system/liquidation.test.js
```diff
@@ -180,7 +181,7 @@ describeFn("System -- liquidation", () => {
       await expect(parseInt(tx.gasLimit)).to.be.lessThan(518000)
 
       const txReceipt = await ethers.provider.getTransactionReceipt(tx.hash)
-      await expect(parseInt(txReceipt.gasUsed)).to.be.lessThan(255000)
+      await expect(parseInt(txReceipt.gasUsed)).to.be.lessThan(400000)
```
pdyraga, line 184: I am surprised by this increase. Can we check on a local geth instance what is the cost? It is possible that the Hardhat does not properly refund.

package.json
```diff
@@ -45,7 +45,7 @@
     "eslint-config-keep": "github:keep-network/eslint-config-keep#0c27ade",
     "ethereum-waffle": "^3.3.0",
     "ethers": "^5.0.32",
-    "hardhat": "^2.1.1",
+    "hardhat": "^2.2.1",
```
pdyraga, line 48: If `yarn.lock` resolves to `2.6.1`, why don't we apply this version here as well?

pdyraga: Oh, I meant `^2.6.1`

tomaszslabon: Done in https://github.com/keep-network/coverage-pools/pull/178/commits/841db95b1bf9cac977338b757d9f576b8d26fe2d.

pdyraga (approved):
I think this is good to go. The gas usage question can be answered separately.

PR #180, opened 2021-08-24
Prevent approving already approved swappers

lukasz-zimnoch:
This change should prevent emitting `SignerBondsSwapperApproved` in case swapper is already approved.

### DIFF
--- a/contracts/SignerBondsUniswapV2.sol
+++ b/contracts/SignerBondsUniswapV2.sol
@@ -261,6 +261,10 @@ contract SignerBondsUniswapV2 is ISignerBondsSwapStrategy, Ownable {
     /// @dev Can be called only by the contract owner.
     /// @param swapper Swapper that will be approved
     function approveSwapper(address swapper) external onlyOwner {
+        require(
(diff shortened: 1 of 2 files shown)


contracts/SignerBondsUniswapV2.sol
```diff
@@ -261,6 +261,10 @@ contract SignerBondsUniswapV2 is ISignerBondsSwapStrategy, Ownable {
     /// @dev Can be called only by the contract owner.
     /// @param swapper Swapper that will be approved
     function approveSwapper(address swapper) external onlyOwner {
+        require(
+            !approvedSwappers[swapper],
+            "Signer bonds swapper has been already approved"
+        );
```
pdyraga, line 267: Can we add a simple unit test to cover this scenario?

lukasz-zimnoch: Done https://github.com/keep-network/coverage-pools/pull/180/commits/0f1425cff6200952c62ecb61d9bca9e878e12b26

pdyraga (approved):

PR #186, opened 2021-08-30
Add workflow deploying contracts on hardhat network

michalinacienciala:
We're adding a workflow which will publish `coverage-pools` contracts
deployed on the `hardhat` network, so they could be later used when
building Token Dashboard on Testnet. The artifacts published by the job
will be used during verification of changes to the Dashboard (when
`Token Dashboard / Testnet` gets triggered by `push` or `pull_request`
event). We're deploying contracts on the `hardhat` network and saving
the artifacts (`--write true`) in the `deployments/hardhat`. The
`prepare-artifacts` script executed during the `npm publish ...` step
will copy the deployed contracts to the `/artifacts` folder.
The addresses of the contracts deployed on `hardhat` network won't
make sense on the Testnet, but that's not a problem, as they're not
used.

The `-dev` suffix will be used in the name of the published package.
The workflow will be triggered every time a change to contracts or
dependencies gets pushed to `main`.

Note: we're temporarily publishing to NPM registry using `--dry-run`
flag. The flag will be removed before merging of the changes
to the `main` branch.


.github/workflows/npm.yml
```diff
@@ -0,0 +1,51 @@
...
+          registry-url: "https://registry.npmjs.org"
+          cache: "yarn"
+
+      - name: Resolve latest contracts
+        run: |
+          yarn add \
+            @keep-network/keep-core \
+            @keep-network/tbtc
+
+      - name: Compile contracts
```
nkuba, line 32:
```suggestion
      # Deploy contracts to a local network to generate deployment artifacts that 
      # are required by dashboard compilation.
      - name: Deploy contracts
```

.github/workflows/npm.yml
```diff
@@ -0,0 +1,51 @@
...
+        run: yarn deploy --network hardhat --write true
+
+      - name: Copy artifacts
+        run: |
+          mkdir -p artifacts
+          cp -r build/contracts/* artifacts/
+
```
nkuba, line 39: We don't need this step in a setup where we use hardhat.
The copying we need will be performed by `npm publish` job with execution of `prepublishOnly` script.

For reference of the package structure please see https://github.com/keep-network/coverage-pools/issues/159.
```suggestion
```

nkuba (changes requested):

.github/workflows/npm.yml
michalinacienciala: Fixed, cc2a2866016ec111b5250e436314f3233bfad061.

.github/workflows/npm.yml
michalinacienciala: Right, I actually thought I removed this one, but apparently not. Fixed in 738b247b5c158e61a77fe697eebca48991e70fcd.

michalinacienciala:
Workflow verified with `--dry-run` flag used in step publishing to NPM registry: https://github.com/keep-network/coverage-pools/pull/186/checks?check_run_id=3461334860. I removed the testing configuration. Ready for the final review.

nkuba (approved):

PR #188, opened 2021-09-02
More efficient master contract check

pdyraga:
Reading immutable field is cheaper since it is stored in the bytecode
and not in the storage.

### DIFF
--- a/contracts/Auction.sol
+++ b/contracts/Auction.sol
@@ -50,7 +50,7 @@ contract Auction is IAuction {
     }
 
     AuctionStorage public self;
-    bool public isMasterContract;
+    address public immutable masterContract;
 
     /// @notice Throws if called by any account other than the auctioneer.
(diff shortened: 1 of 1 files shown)


tomaszslabon:
@pdyraga 
> Reading immutable field is cheaper since it is stored in the bytecode
> and not in the storage.

I ran all our unit tests to check how changes on this branch affect functions that check if the `auction` is the `master auction`:
- `notifyLiquidation`:
  - `main`: 253737
  - `more-tweaks`: 252605
    
- `notifyLiquidated`:
  - `main`: 83316
   - `more-tweaks`: 81232
     
- `takeOffer`:
  - `main`: 104276
  - `more-tweaks`: 103558
  
 This code is less readable, but uses slightly less gas, at least in our unit tests.

dimpar (approved):
lgtm, wrapping the check into `isMasterContract()` helps with readability.

PR #189, opened 2021-09-02
Validate auction length is non-zero when construction auction

pdyraga:
Depends on #188 

We are validating amount desired so we should also validate auction
length.


tomaszslabon:
> Depends on #188
> 
> We are validating amount desired so we should also validate auction
> length.

@pdyraga This PR makes sense, I think we can merge it.

dimpar (approved):
lgtm

PR #195, opened 2021-09-17
Verify contracts on Etherscan

michalinacienciala:
We're adding a step verifying the migrated contracts on Etherscan. We're
using `etherscan-verify` plugin. In order for the plugin not to pick up
wrong files for verification we have to remove contracts of the
dependent projects stored in the `node_modules`.

We decided to move the etherscan verification to a separate job,
executed after the job which migrates contracts. This way the failure
of verification will not block anything and will be easily visible as a red
status dot.
To have the verification in a separate job, we need to pass the migrated
contracts between the jobs. I considered also passing all needed
`node-modules` subfolders, to get rid of the `yarn install` step, but
the number of needed modules was big, which would make maintaining
the workflow file troublesome.

Refs:
https://github.com/keep-network/keep-core/pull/2598
https://github.com/keep-network/keep-ecdsa/pull/913
https://github.com/keep-network/tbtc/pull/834
https://github.com/keep-network/tbtc-v2/pull/52


michalinacienciala:
Workflow run proving successful contracts verification: https://github.com/keep-network/coverage-pools/runs/3633195118?check_suite_focus=true
Workflow run showing annotations added in case of verification failure: https://github.com/keep-network/tbtc-v2/actions/runs/1245161578 (in `tbtc-v2` project, same implementation of the annotation action)

michalinacienciala:
Initial implementation executed etherscan verification in the same job
as migration and utilized annotations to inform about failures in the
verification process.
We decided to move the etherscan verification to a separate job,
executed after the job which migrates contracts. This way the failure
of verification will not block anything and will be easily visible as a red
status dot.
To have the verification in a separate job, we need to pass the migrated
contracts between the jobs. I considered also passing all needed
node-modules subfolders, to get rid of the yarn install step, but
the number of needed modules was big, which would make maintaining
the workflow file troublesome.

I've tested the new solution on seperate branch, in following workflow:
https://github.com/keep-network/coverage-pools/runs/3729852018?check_suite_focus=true.

nkuba (approved):

PR #199, opened 2021-10-14
Add governance support for T coverage pool underwriter token holders

tomaszslabon:
This PR extends the `UnderwriterToken` by Compound-like voting and delegation features.

The following changes were made:
- updated `thesis/solidity-contracts` to the newest version
- added dependency `threshold-network/solidity-contracts`
- made `UnderwriterToken` be derived from `Checkpoints` (implemented functions `delegate()` and `delegateBySig()`)
- overridden function `beforeTokenTransfer` so that voting power is given on minting and burning tokens.

 `Checkpoints` can only work with up to `2^96 - 1` tokens. To ensure this requirement is handled properly, there is a check that prevents minting more `UnderwriterTokens` than 2^96-1.

### DIFF
--- a/test/AssetPool.test.js
+++ b/test/AssetPool.test.js
@@ -385,6 +385,29 @@ describe("AssetPool", () => {
     )
 
     context(
+      "when the depositor deposits maximum allowed amount of tokens",
+      () => {
+        beforeEach(async () => {
+          const amount = MAX_UINT96
+          const minCovToMint = amount
+          await collateralToken.mint(underwriter1.address, amount)
+          await collateralToken
+            .connect(underwriter1)
+            .approve(assetPool.address, amount)
+          await assetPool
+            .connect(underwriter1)
+            .depositWithMin(amount, minCovToMint)
+        })
+
+        it("should mint the right amount of underwriter tokens", async () => {
+          expect(
+            await underwriterToken.balanceOf(underwriter1.address)
+          ).to.equal(MAX_UINT96)
+        })
+      }
+    )
+
+    context(
       "when amount to be minted is smaller than the minimal required",
       () => {
         const depositedUnderwriter1 = to1e18(100)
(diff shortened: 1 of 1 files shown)


contracts/UnderwriterToken.sol
```diff
@@ -27,8 +28,89 @@ import "@thesis/solidity-contracts/contracts/token/ERC20WithPermit.sol";
 ///         Anyone can submit this signature on the user's behalf by calling the
 ///         permit function, as specified in EIP2612 standard, paying gas fees,
 ///         and possibly performing other actions in the same transaction.
-contract UnderwriterToken is ERC20WithPermit {
+contract UnderwriterToken is ERC20WithPermit, Checkpoints {
+    /// @notice Replay protection nonces that are used to construct EIP-712
+    ///         signature provided to `delegateBySig` function.
+    mapping(address => uint256) public delegationNonce;
```
pdyraga, line 34: Why we do not reuse `nonce` from `ERC20WithPermit`? I see we reuse it in T token implementation [here](https://github.com/threshold-network/solidity-contracts/blob/6664c738660f79de3add7fdff735fcb19d5165ad/contracts/token/T.sol#L61).

contracts/UnderwriterToken.sol
```diff
@@ -27,8 +28,89 @@ import "@thesis/solidity-contracts/contracts/token/ERC20WithPermit.sol";
 ///         Anyone can submit this signature on the user's behalf by calling the
 ///         permit function, as specified in EIP2612 standard, paying gas fees,
 ///         and possibly performing other actions in the same transaction.
-contract UnderwriterToken is ERC20WithPermit {
+contract UnderwriterToken is ERC20WithPermit, Checkpoints {
```
pdyraga, line 31: It also needs to override `beforeTokenTransfer` to write voting power and total supply checkpoints in case tokens were minted/burned.

PR #202, opened 2021-10-29
T cov pool: Do not allow the total value of collateral locked to exceed 2^96 - 1

pdyraga:
In the previous PRs (see #199) we limited the amount of collateral locked to `2^96 - 1` and added support for governance checkpoints to `UnderwriterToken` contract.

Unfortunately, the validation in deposit function was not entirely correct. Instead of validating that a single deposit can not exceed `2^96 - 1`, we should expect that the total value locked in the pool, including all previous deposits plus the amount that is about to be deposited, does not exceed `2^96 - 1`. This way we better guard that no more than `2^96 - 1` underwriter token is minted. 

With the previous version, it was possible to perform multiple deposits and bypass the validation check. This would still fail later on `UnderWriterToken`'s `beforeTokenTransfer` check but the validation checks in deposit functions were not entirely correct and intuitive.

### DIFF
--- a/contracts/AssetPool.sol
+++ b/contracts/AssetPool.sol
@@ -583,16 +583,6 @@ contract AssetPool is Ownable, IAssetPool {
         uint256 tokensToMint = (amountToDeposit * covSupply) /
             collateralBalance;
 
-        // The total supply of underwriter tokens cannot exceed `type(uint96).max`,
-        // because the maximum voting power in the underwriter token is
-        // `type(uint96).max`. Therefore the asset pool should not allow to mint
-        // more than `type(uint96).max` either. The amount of tokens to mint is
-        // stored in `uint256` just for gas efficiency.
-        require(
-            tokensToMint <= type(uint96).max,
-            "Minted tokens amount must be <= 2^96 - 1"
-        );
-
         return tokensToMint;
     }
 
--- a/contracts/AssetPool.sol
+++ b/contracts/AssetPool.sol
@@ -146,7 +146,8 @@ contract AssetPool is Ownable, IAssetPool {
 
     /// @notice Accepts the given amount of collateral token as a deposit and
     ///         mints underwriter tokens representing pool's ownership. The
-    ///         amount must be smaller or equal to 2^96-1.
+    ///         amount locked in the pool after accepting this deposit must
+    ///         be smaller or equal to 2^96-1; otherwise, the function reverts.
     ///         Optional data in extraData may include a minimal amount of
     ///         underwriter tokens expected to be minted for a depositor. There
     ///         are cases when an amount of minted tokens matters for a
@@ -159,8 +160,8 @@ contract AssetPool is Ownable, IAssetPool {
         bytes calldata extraData
     ) external {
         require(
-            amount <= type(uint96).max,
-            "deposited amount must be <= 2^96 - 1"
+            amount + totalValue() <= type(uint96).max,
+            "Pool capacity exceeded"
         );
         require(msg.sender == token, "Only token caller allowed");
         require(
@@ -182,7 +183,9 @@ contract AssetPool is Ownable, IAssetPool {
     }
 
     /// @notice Accepts the given amount of collateral token as a deposit and
-    ///         mints underwriter tokens representing pool's ownership.
+    ///         mints underwriter tokens representing pool's ownership. The
+    ///         amount locked in the pool after accepting this deposit must
+    ///         be smaller or equal to 2^96-1; otherwise, the function reverts.
     /// @dev Before calling this function, collateral token needs to have the
     ///      required amount accepted to transfer to the asset pool.
     /// @param amountToDeposit Collateral tokens amount that a user deposits to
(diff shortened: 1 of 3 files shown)


contracts/AssetPool.sol
```diff
@@ -195,8 +198,8 @@ contract AssetPool is Ownable, IAssetPool {
         returns (uint256)
     {
         require(
-            amountToDeposit <= type(uint96).max,
-            "deposited amount must be <= 2^96 - 1"
+            amountToDeposit + totalValue() <= type(uint96).max,
```
dimpar, line 201: Shouldn't we check `toMint + totalValue() <= type(uint96).max` instead? My understanding is that `amountToDeposit` not necessary equals to `toMint`. But `totalValue()` checks already minted tokens, no?

dimpar:
We should probably update [this](https://github.com/keep-network/coverage-pools/blob/main/contracts/AssetPool.sol#L587) code. If [my](https://github.com/keep-network/coverage-pools/pull/202#discussion_r740252301) comment makes sense, maybe we can add `tokensToMint + totalValue() <= type(uint96).max` here and remove all 3 other `require` checks?

contracts/AssetPool.sol
pdyraga: Let me explain the whole story. `type(uint96).max` is a limitation of [`Checkpoint`](https://github.com/threshold-network/solidity-contracts/blob/main/contracts/governance/Checkpoints.sol) contract we implemented for T Governance. Given that we will never mint more than `2^96 - 1` T, we could save some gas for every token operation (including transfers) by tightly packing values and restricting the voting power to `2^96 - 1`.

Given that we want coverage pool underwriters to participate in DAO, we need to have `UnderwriterToken` implement `Checkpoints`. It means that we need to limit the total supply of `UnderwriterToken` to `type(uint96).max` and we do it [here](https://github.com/keep-network/coverage-pools/blob/0ab5314fceb8e3849a13aa6f9918c14be11ce76a/contracts/UnderwriterToken.sol#L117-L120). That check should be enough to ensure the correctness of operations though it felt wrong to not clearly express this limitation on the `AssetPool` level as well. Even if it happens there is more T minted than `2^96 - 1`, the coverage pool could still work by just allowing to deposit up to `2^96 - 1` in the happy path. And this is what we are covering but the three checks here. 

With these three checks, in a very unlikely scenario when more than `2^96 - 1` T is ever minted, when the pool reaches its max capacity and someone tries to deposit more tokens, the revert message will be "Pool capacity exceeded" from `AssetPool` instead of quite mysterious "Maximum total supply exceeded" from `UnderwriterToken`. 

It may still happen we will get "Maximum total supply exceeded" in case there was a major liquidation and there is some tiny fraction of T left in the pool. This can be easily fixed by donating e.g. 1 T with a plain transfer. The same problem is for `uint256` version though.

If we are looking for optimizations, [this code](https://github.com/keep-network/coverage-pools/blob/e85b7d624368eaf38c0405ae6b225e3be3bcc4d9/contracts/AssetPool.sol#L581-L589) could probably be removed altogether because it is [already checked](https://github.com/keep-network/coverage-pools/blob/0ab5314fceb8e3849a13aa6f9918c14be11ce76a/contracts/UnderwriterToken.sol#L117-L120). I am fine with keeping and removing it. Having this limitation clearly expressed on the `AssetPool` level makes sense though. You can not deposit more and you can not mint more because minting is 1:1 in the happy path.

dimpar: Ok, it makes sense to leave these 3 checks. If we want to leave [this check in _calculateTokensToMin](https://github.com/keep-network/coverage-pools/blob/e85b7d624368eaf38c0405ae6b225e3be3bcc4d9/contracts/AssetPool.sol#L581-L589) then shouldn't we modify it to:
```
require(
            tokensToMint <= type(uint96).max - covSupply,
            "Minted tokens amount must be <= 2^96 - 1"
);
```
? Unless we just remove it to save a bit of gas since [other](https://github.com/keep-network/coverage-pools/blob/0ab5314fceb8e3849a13aa6f9918c14be11ce76a/contracts/UnderwriterToken.sol#L117-L120) `require` will validate the same thing. Up to you.

pdyraga: Okay, let's do the following:
- Remove `tokensToMint <= type(uint96).max` in `AssetPool` because it currently does not make sense - your last comment, and rely on the check in `UnderwriterToken.beforeTokenTransfer`
- Let's leave a unit test for `AssetPool` ensuring the transaction fails on `UnderwriterToken.beforeTokenTransfer`

Also, let's simplify the logic in `UnderwriterToken.beforeTokenTransfer`. We do not need the `SafeCast` at the beginning because if we validate the maximum supply in minting, there is no way to exceed it when burning or transferring.

pdyraga: https://github.com/keep-network/coverage-pools/pull/202/commits/7b3c01d396d51c903d81e9a04379e3a1ede89f91

contracts/AssetPool.sol
```diff
@@ -578,16 +583,6 @@ contract AssetPool is Ownable, IAssetPool {
         uint256 tokensToMint = (amountToDeposit * covSupply) /
```
dimpar, line 583: now we can just: `return (amountToDeposit * covSupply) / collateralBalance`

dimpar:
Other changes look good to me

contracts/AssetPool.sol
pdyraga: Right, I'll fix it.

pdyraga: ada4c54ebda5ebefc65f5a9a9b26ef04440a5971 sir

dimpar (approved):

Issue #200, opened 2021-10-21
T Coverage Pool deployment


dimpar:
Coverage Pool migration proposal: https://forum.threshold.network/t/tip-043-coverage-pool-migration/465

Team | Address
------------ | -------------
**Deployment Team** | `0x123694886DBf5Ac94DDA07135349534536D14cAf`
**Threshold Council** | `0x9f6e831c8f8939dc0c830c6e492e7cef4f9c2f5f`
**Treasury Guild** | `0x71E47a4429d35827e0312AA13162197C23287546`

# Prerequisites
- [x] https://github.com/keep-network/coverage-pools/pull/227
- [x] https://github.com/keep-network/coverage-pools/issues/220

# Deploy smart contracts
All steps done by the Deployment Team
- [x] Create release branch
- [x] Update dependencies to mainnet ones
- [x] Update version to `2.0.0`
- [x] Tag `solidity/v2.0.0` in `keep-network/coverage-pools`
- [x] Sign `solidity/v2.0.0` commit hash
- [x] Deploy `UnderwriterToken` contract (address: `0x007854FdCC9c6F748b435e63745599fb1A0dF843`) with the following parameters:
  - `_name`: `covT underwriter token`
  - `_symbol`: `covT`  
- [x] Deploy `AssetPool` (address: `0x21b86FDC93F23fD52022A283828dF8E808D28D5D`) with the following parameters:
  - `_collateralToken`: `T@0xCdF7028ceAB81fA0C6971208e83fa7872994beE5`
  - `_underwrtierToken`: `UnderwriterToken.address` (deployed in previous step)
  - `rewardsManager`: Deployment Team (temporarily, will be transferred in next steps)
- [x] Transfer `UnderwriterToken` ownership to `AssetPool` (TX: `0xefa3052c96fb054bfd5a377e81b3cabe525e6ddd20937d33200a6f1fcd70c4aa`)
- [x] Deploy `CoveragePool` (address: `0x7D7B622C07f216d99eFc3EfB9Cf71872F46D811A`) with the following parameters:
  - `_assetPool`: `AssetPool.address` (deployed in previous step)
- [x] Transfer `AssetPool` ownership to `CoveragePool` (TX: `0xf4517bfba17413235a8e73c4db4f2d4aaa3a0476f21a49db4983194c2d536f0e`)
[...]

PR #203, opened 2021-11-01
Updating deployment scripts for T cov pool

dimpar:
Refs https://keepnetwork.atlassian.net/browse/CP-134
Deployment scenario: https://github.com/keep-network/coverage-pools/issues/200

This PR adds changes in T coverage pool deployment scripts. Most of the changes were around
- Replacing KeepToken with `T` token
- Removing deployment scripts related to coverage pool v1 (RiskManager, Auctions etc.)
- Updating abi
- Introducing `thresholdCouncil` and `treasuryGuild` accounts
- Updating solidity version


deploy/00_resolve_t_token.ts
```diff
@@ -6,19 +6,19 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
   const { log } = deployments
   const { deployer } = await getNamedAccounts()
 
-  const KeepToken = await deployments.getOrNull("KeepToken")
+  const TToken = await deployments.getOrNull("TToken")
```
pdyraga, line 9: Should it be `TToken` or `T`? I am looking at `threshold-network/solidity-contracts` [deployment script of the token staking contract](https://github.com/threshold-network/solidity-contracts/blob/4b363a6803389d0a83dc295e25218ee03ce8c925/deploy/07_deploy_token_staking.ts#L11) and T token is referenced by `const T = await deployments.get("T")` there. Also, manner T token artifacts are saved in [`T.json`](https://github.com/threshold-network/solidity-contracts/blob/main/deployments/mainnet/T.json).

deploy/03_resolve_uniswap_v2_router.ts
```diff
@@ -1,6 +1,8 @@
 import { HardhatRuntimeEnvironment, HardhatNetworkConfig } from "hardhat/types"
 import { DeployFunction } from "hardhat-deploy/types"
 
+// TODO: THIS DEPLOYMENT CAN BE DONE ONLY IF THERE IS A UNISWAP POOL WITH
```
pdyraga, line 4: Let's drop all deployment scripts related to v1 risk manager. T coverage pool will not support this way of claiming the coverage, the only option will be the multisig claim implemented in #216. Details in this [proposal](https://forum.threshold.network/t/tip-043-coverage-pool-migration/465). 

Let's also add `@dev` comment to v1 risk-manager-related contracts that the given contract is used only for KEEP coverage pool in v1 that got sunset, and drop a link to TIP-043.

external/mainnet/TToken.json
```diff
@@ -0,0 +1,4 @@
+{
```
pdyraga, line 1: Here it is: https://github.com/threshold-network/solidity-contracts/blob/main/deployments/mainnet/T.json

external/mainnet/UniswapV2Router.json
```diff
@@ -1,976 +1,4 @@
 {
```
pdyraga, line 1: Instead of deleting this file or `KeepToken.json`, I propose we rename `external/mainnet` to `external/mainnet-v1` and keep the files unchanged there. We can then have `external/mainnet` for v2 or `externa/mainnet-v2`. IMO second option is cleaner.

deploy/11_deploy_risk_manager_v1.ts
```diff
@@ -10,6 +10,8 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
   const TBTCDepositToken = await deployments.get("TBTCDepositToken")
   const CoveragePool = await deployments.get("CoveragePool")
   const SignerBondsManualSwap = await deployments.get("SignerBondsManualSwap")
+  // TODO: Determine if the ETH<->T pool exists. If there's no pool, then
+  //       this strategy should not be deployed.
```
pdyraga, line 14: No, everything is happening manually via the proposal. Let's drop it.

deploy/12_deploy_coverage_pool_beneficiary.ts
```diff
@@ -6,7 +6,7 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
   const { read } = deployments
```
pdyraga, line 6: We'll have to confirm how the Treasury Guild wants to send rewards and if they want to do it via escrow. I suspect the answer is yes because they will probably mint some amount of tokens, fund the escrow, and then release rewards every week or so.

pdyraga (commented):
The first round of comments, I'll work on updating https://github.com/keep-network/coverage-pools/issues/200 in the meantime.

external/mainnet/UniswapV2Router.json
dimpar: Done here, but why do we need to keep `mainnet-v1`? In this PR we are removing a bunch of deployment scripts related to v1, and I take that some of the code will also be changed/removed in the next PRs. Maybe we should remove it and just leave `mainnet`?

deploy/00_resolve_t_token.ts
dimpar: Done [6f2a9df](https://github.com/keep-network/coverage-pools/pull/203/commits/6f2a9dfef1239d00a8ff089c9069538cb92aaaed)

deploy/03_resolve_uniswap_v2_router.ts
dimpar: Partially done in [82173bc](https://github.com/keep-network/coverage-pools/pull/203/commits/82173bc726c4622bda153c68c16c9aade559484e)

external/mainnet/TToken.json
dimpar: Added as part of [82173bc](https://github.com/keep-network/coverage-pools/pull/203/commits/82173bc726c4622bda153c68c16c9aade559484e)

deploy/11_deploy_risk_manager_v1.ts
dimpar: I removed the code related to the Risk Manager v1

external/mainnet/UniswapV2Router.json
pdyraga: For `@keep-network/keep-core` we decided not to remove contracts but to archive them into [`solidity-v1` directory](https://github.com/keep-network/keep-core/tree/main/solidity-v1). The reasoning was that these contracts are deployed to mainnet and still in use. The same situation is here except that it would be harder to archive them, given we are going to use part of the contracts for v2.  That's why I [suggested](https://github.com/keep-network/coverage-pools/pull/203#discussion_r1052028944) marking contracts that are used only in v1 in a `@dev` comment.

Now, we have a question what to do with v1 mainnet deployment artifacts. If we are not deleting contracts but archiving them, I think it makes sense to keep artifacts as well. That's just my personal preference so I can see us relying only on the version control but... this wouldn't be consistent with the decision of archiving contracts IMO.

cc @nkuba for thoughts here

.github/workflows/contracts.yml
```diff
@@ -211,7 +211,7 @@ jobs:
 
       - name: Install Solidity
         env:
-          SOLC_VERSION: 0.8.5 # according to solidity.version in hardhat.config.js
+          SOLC_VERSION: 0.8.9 # according to solidity.version in hardhat.config.ts
```
pdyraga, line 214: `@threshold-network/solidity-contracts` uses `0.8.9` but we should be fine even with the most recent version, `0.8.17`. This is the version [used by `tbtc-v2`](https://github.com/keep-network/tbtc-v2/blob/f1d68c10a4b2276959e44d006b3ebd5c09da9876/solidity/contracts/bridge/Bridge.sol#L16).

hardhat.config.ts
```diff
@@ -88,7 +88,7 @@ const config: HardhatUserConfig = {
         "node_modules/@keep-network/tbtc/artifacts",
         "./external/ropsten",
       ],
-      mainnet: ["./external/mainnet"],
+      mainnet: ["./external/mainnet-2"],
```
pdyraga, line 91: `mainnet-2` or `mainnet-v2`?

.github/workflows/contracts.yml
pdyraga: Worth noting, we should also bump pragma in `*.sol` files to make it work as expected.

dimpar: hm, the most recent version `0.8.17` didn't work out of the box even when I added multiple compilers in the hardhat.
```
Error HH606: The project cannot be compiled, see reasons below.
These files import other files that use a different and incompatible version of Solidity:
(here are a bunch of contracts listing different solidity versions 0.8.9 vs 0.8.17)
```

deploy/03_resolve_uniswap_v2_router.ts
dimpar: Added a dev note only for one contract. If the wording looks good to you, I'll add it in other contracts. [7b22857](https://github.com/keep-network/coverage-pools/pull/203/commits/7b228572426d60aa5c4dd61ca1a19987baed676b)

hardhat.config.ts
dimpar: [3ddf2c3](https://github.com/keep-network/coverage-pools/pull/203/commits/3ddf2c344edb7f1e5bc9433fa6d6b89c1b86ff43)

.github/workflows/contracts.yml
pdyraga: I see. We would have to update contracts in https://github.com/threshold-network/solidity-contracts/tree/main/contracts/governance to use `pragma solidity ^0.8.9;`. Not something we have to worry about now.

deploy/12_deploy_coverage_pool_beneficiary.ts
pdyraga: Captured in #220

contracts/Auction.sol
```diff
@@ -30,7 +30,11 @@ import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
 ///         pool. Auctions can be partially filled, and are meant to be amenable
 ///         to flash loans and other atomic constructions to take advantage of
 ///         arbitrage opportunities within a single block.
-/// @dev  Auction contracts are not meant to be deployed directly, and are
+/// @dev  This contract is used only for KEEP coverage pool in v1 that was
+///       depricated. Going forward T coverage pool will be used for v2.
```
pdyraga, line 34:
```suggestion
///       deprecated. Going forward T coverage pool will be used for v2.
```

pdyraga (approved):

nkuba (commented):
I have no idea why I added review comments but have never submitted them... Better late than never 😜

PR #205, opened 2022-01-06
Add environment tag to the published npm package

michalinacienciala:
This change adds an additional environment tag for packages published
during the `Publish to npm` step. This adds a convenient reference to
the latest versions of the `tbtc` package for each environment.


r-czajkowski (approved):
LGTM! :shipit:

PR #216, opened 2022-08-19
Multisig claim

dimpar:
This PR enables council multisig (which was selected by DAO) to claim coverage from the T coverage pool if needed. Only the council multisig can claim a specific amount of funds by calling `RiskManagerV2::claimCoverage(amountToSeize)`.


contracts/RiskManagerV1.sol
```diff
@@ -139,6 +139,9 @@ contract RiskManagerV1 is IRiskManagerV1, Auctioneer, Ownable {
     ///         assuming enough surplus will accumulate up to that point.
     uint256 public tbtcSurplus;
 
+    /// @notice TODO: add description
+    address public councilMultisig;
```
pdyraga, line 143: I would introduce another RiskManager contract for that. As far as I recall, the coverage pool can have multiple risk managers approved.

Also, T coverage pool will not cover TBTC v1 liquidations.

contracts/RiskManagerV1.sol
```diff
@@ -517,6 +535,19 @@ contract RiskManagerV1 is IRiskManagerV1, Auctioneer, Ownable {
         signerBondsSwapStrategyInitiated = 0;
     }
 
+    /// TODO: add description
+    ///      - do we need a delay "onlyAfterGovernanceDelay" for updating council
+    ///        multisig?
```
pdyraga, line 540: Could be worth it, yes. But this logic can be a part of the risk manager.

contracts/RiskManagerV1.sol
dimpar: Yes, the coverage pool can have multiple risk managers approved. Added a new `RiskManagerV2`.

contracts/RiskManagerV1.sol
dimpar: Done

anishnaik:
### (TOB-PR216-1) Missing zero-address checks

**Severity**: Informational
**Difficulty**: High

During the construction of a `RiskManagerV2` contract, there is no zero-address check for the `tbtcToken` address. 

https://github.com/keep-network/coverage-pools/blob/62c117d855fc1d60cbf7556768c8d078c9aaeef8/contracts/RiskManagerV2.sol#L86

Setting this address to the zero-address, by mistake, may lead to undefined behavior.

**Recommendations**

- [ ] Short term, add a zero-address check before setting the `tbtcToken` address.
- [ ] Long term, improve unit test coverage to find edge cases similar to this one.

anishnaik:
### (TOB-PR216-2) Insufficient post-condition checks for unit tests

**Severity**: Informational
**Difficulty**: N/A

The unit tests created for this PR do not sufficiently test the state of contracts after an action is taken. This applies to the following 2 scenarios:
1. `beginCouncilMultisigUpdate` tests: The `newCouncilMultisig` address should be checked to make sure that it is equal to `_newCouncilMultisig`, assuming that the `owner` and not `thirdParty` calls the function. Additionally, another test should be added to check that setting `_newCouncilMultisig` to `address(0)` causes the transaction to revert.
2. `finalizeCouncilMultisigUpdate` tests:  The `newCouncilMultisig` value should be checked to make sure that it is set to `address(0)`.

**Recommendations**

- [ ] Short term, add the post-condition checks mentioned above to the unit tests for this PR.
- [ ] Long term, ensure that unit tests check the values of all state variables that are impacted by the action performed in the unit test.

anishnaik:
### Code Coverage

We performed a security review of this PR for a total of 8 engineer hours.  Here is a non-exhaustive list of our approaches taken during this review.

1. Are the correct access controls in place to prevent anyone from seizing funds?
2. Are the correct access controls in place to prevent anyone from becoming the next `councilMultisig`?
3. Can the `GOVERNANCE_DELAY` be bypassed?
4. Can `finalizeCouncilMultisigUpdate` be called more than once to cause undefined behavior?
5. Is the `councilMultisigInitiated` value updated correctly across the various functions?
6. Does `getRemainingCouncilMultisigUpdateTime` correctly track the remaining time before the `newCouncilMultisig` can be finalized?
7. Do the unit tests sufficiently cover all possible edge cases? (see [TOB-PR216-2](https://github.com/keep-network/coverage-pools/pull/216#issuecomment-1309380268)).
8. Are there any front-running opportunities?

dimpar:
> d719ec9

Agree. Addressed in [d719ec9](https://github.com/keep-network/coverage-pools/pull/216/commits/d719ec9fd62377dde543a774f7878cd0b1efaa76)

dimpar:
> ### (TOB-PR216-1) Missing zero-address checks
> **Severity**: Informational **Difficulty**: High
> 
> During the construction of a `RiskManagerV2` contract, there is no zero-address check for the `tbtcToken` address.
> 
> https://github.com/keep-network/coverage-pools/blob/62c117d855fc1d60cbf7556768c8d078c9aaeef8/contracts/RiskManagerV2.sol#L86
> 
> Setting this address to the zero-address, by mistake, may lead to undefined behavior.
> 
> **Recommendations**
> 
> * [ ]  Short term, add a zero-address check before setting the `tbtcToken` address.
> * [ ]  Long term, improve unit test coverage to find edge cases similar to this one.

Agree, addressed in [d719ec9](https://github.com/keep-network/coverage-pools/pull/216/commits/d719ec9fd62377dde543a774f7878cd0b1efaa76)

dimpar:
> ### (TOB-PR216-2) Insufficient post-condition checks for unit tests
> **Severity**: Informational **Difficulty**: N/A
> 
> The unit tests created for this PR do not sufficiently test the state of contracts after an action is taken. This applies to the following 2 scenarios:
> 
> 1. `beginCouncilMultisigUpdate` tests: The `newCouncilMultisig` address should be checked to make sure that it is equal to `_newCouncilMultisig`, assuming that the `owner` and not `thirdParty` calls the function. Additionally, another test should be added to check that setting `_newCouncilMultisig` to `address(0)` causes the transaction to revert.
> 2. `finalizeCouncilMultisigUpdate` tests:  The `newCouncilMultisig` value should be checked to make sure that it is set to `address(0)`.
> 
> **Recommendations**
> 
> * [ ]  Short term, add the post-condition checks mentioned above to the unit tests for this PR.
> * [ ]  Long term, ensure that unit tests check the values of all state variables that are impacted by the action performed in the unit test.

Agree. Addressed in [55735ab](https://github.com/keep-network/coverage-pools/pull/216/commits/55735abf18e86f382e26e8eb0682c13ae5b73eff)

contracts/CoveragePool.sol
```diff
@@ -223,6 +224,26 @@ contract CoveragePool is Ownable {
...
 
+    /// @notice Seizes funds from the coverage pool and puts them aside for the
+    ///         recipient to withdraw. This function in contrary to the other
+    ///         'seizeFunds' accepts a specific amount to be seized from the pool
+    ///         but no more than the pool holds.
+    /// @dev    Note the order of params compared to the other 'seizeFunds'.
+    ///         Here we seize a specific amount of funds.
+    /// @param  amount Amount to be seized from the pool.
+    /// @param  recipient Address that will receive the pool's seized funds.
+    function seizeFunds(uint256 amount, address recipient)
```
pdyraga, line 235: To lower the possibility of a human error we could have the same order of params but use different names for functions:
- `seizePortionOfPool(address recipient, uint256 portionToSeize)`
- `seizeAmountFromPool(address recipient, uint256 amount)`

contracts/CoveragePool.sol
```diff
@@ -223,6 +224,26 @@ contract CoveragePool is Ownable {
...
+        external
+        onlyApprovedRiskManager
+    {
+        require(
+            amount <= collateralToken.balanceOf(address(assetPool)),
+            "Amount to seize exceeds the pool balance"
+        );
```
pdyraga, line 242: Wouldn't it make more sense to have this validation in `AssetPool.claim`? It feels that it's the `AssetPool`'s responsibility to validate its own balance when `claim` with the given `amount` is called.

contracts/RiskManagerV2.sol
```diff
@@ -0,0 +1,158 @@
...
+///         coverage from the coverage pool. Demanding coverage is akin to
+///         filing a claim in traditional insurance and processing your own
+///         claim. The risk manager holds an incredibly privileged position,
+///         because the ability to claim coverage of an arbitrarily large
+///         position could bankrupt the coverage pool.
+///         tBTC v2 risk manager demands coverage by opening an auction for TBTC
+///         and liquidating portion of the coverage pool when tBTC v2 deposit is
+///         in liquidation and signer bonds on offer reached the specific
+///         threshold. In practice, it means no one is willing to purchase
+///         signer bonds for that deposit on tBTC side.
```
pdyraga, line 37: I think we had a communication glitch. For v2, we do not want to run auctions. It's as simple as having some multisig being able to claim the coverage. From [TIP-043](https://forum.threshold.network/t/tip-043-coverage-pool-migration/465) proposal:

> - Treasury Guild becomes risk manager on the T-only Coverage Pool. 
> - This is the group responsible for claiming coverage out of the pool if needed. 
> - The Treasury Guild has autonomy over how this role operators within the DAO and can make changes when needed.
> - The DAOs funds in T are at risk only so the can call coverage from pool without limit
> - Move fast in event where coverage needs to be claimed. Not time for full DAO ratification
> - Further discussion on an extended risk manager role is discussed in [this forum thread](https://forum.threshold.network/t/risk-manager-for-tbtcv2/463). This content of this discussion was determined to be outside of the scope of this proposal

That being said, I do not think we need `RiskManagerV2` at all. If the Treasury Guild multisig is an approved risk manager in `CoveragePool`, they can call `seizeFunds` (or `seizePortionOfPool` / `seizeAmountFromPool` directly.

contracts/interfaces/IRiskManagerV2.sol
```diff
@@ -0,0 +1,33 @@
+// ██████████████     ▐████▌     ██████████████
```
pdyraga, line 1: Let's drop it if it stays empty. And I believe it does based on the previous comment.

pdyraga:
Replaced by https://github.com/keep-network/coverage-pools/pull/223.

PR #231, opened 2023-01-11
Adjust CI to support deployments of contracts on Goerli testnet

michalinacienciala:
Previously we've been deploying coverage-pools on Ropsten. As this testnet
became deprecated, we're switching to the Goerli testnet. The hardhat config for
that network has already been partially added in a separate commit/PR. Now we
need to adjust CI jobs and add missing pieces in the Hardhat config to make the
deployment work.

As part of this PR:
- we're adjusting the names of the environment variables and GH secrets used in the workflow
- we're incorporating execution of the Solidity workflow in the CI flow (to make it work we need https://github.com/keep-network/ci/pull/35 merged)
- we're adjusting Hardhat and Tenderly config to support Goerli
- we're removing obsolete Ropsten configuration
- we're adding goerli UniswapV2Router artifact
- we're doing clean up of the way we verify contracts on Etherscan

### DIFF
--- a/.github/workflows/contracts.yml
+++ b/.github/workflows/contracts.yml
@@ -121,6 +121,7 @@ jobs:
         env:
           CHAIN_API_URL: ${{ secrets.GOERLI_ETH_HOSTNAME_HTTP }}
           ACCOUNTS_PRIVATE_KEYS: ${{ secrets.GOERLI_ETH_CONTRACT_OWNER_PRIVATE_KEY }}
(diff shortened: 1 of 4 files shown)


michalinacienciala:
@nkuba, after https://github.com/keep-network/ci/pull/35 gets merged, I'll update the `v2` tag there and will create a temporary commit here that will allow for the test of the changes included in this PR. If tests will be green, I'll revert that temp commit and mark this PR as ready for review. But you can start the review even now.

